PR #132, opened 2026-04-02
[KMP] Add static Android Typeface accessor for LemonadeTextStyle

bruno-vsiqueira:
## Summary
- Adds `LemonadeTextStyle.typeface(context: Context): Typeface` extension for Android
- Mirrors iOS's `LemonadeTextStyle.uiFont` property, enabling native SDK integrations to use Lemonade fonts without Compose context
- Font bytes are loaded from `LemonadeRes.font.*` (generated compose resources), so no hardcoded asset paths
- Typefaces are cached per font weight after first load

## Motivation
Native Android SDKs (e.g. MeaCardData for card data display) require `android.graphics.Typeface`, which currently can only be resolved from Compose context via `lemonadeFontFamily` + `LocalFontFamilyResolver`. This forces consumers to create SDK providers at the UI layer instead of at DI time, creating asymmetry with iOS where `LemonadeTypography.shared.bodyMediumSemiBold.uiFont` works anywhere.

## Usage
```kotlin
val style = LemonadeTypography.BodyMediumSemiBold.style
val typeface = style.typeface(context) // android.graphics.Typeface
val fontSize = style.fontSize          // Float (sp)
```

## Test plan
- [ ] Verify `compileReleaseKotlinAndroid` passes
- [ ] Verify typeface is correctly loaded (Figtree-SemiBold) at runtime
- [ ] Verify caching works (second call returns cached instance)

### DIFF
--- a/kmp/ui/src/androidMain/kotlin/com/teya/lemonade/LemonadeTypefaces.kt
+++ b/kmp/ui/src/androidMain/kotlin/com/teya/lemonade/LemonadeTypefaces.kt
@@ -30,13 +30,17 @@ public fun LemonadeTextStyle.typeface(context: Context): Typeface {
     }
 }
 
-private fun Int.toFontResource(): FontResource = when (this) {
-    500 -> LemonadeRes.font.Figtree_Medium
-    600, 700 -> LemonadeRes.font.Figtree_SemiBold
-    else -> LemonadeRes.font.Figtree_Regular
-}
+private fun Int.toFontResource(): FontResource =
+    when (this) {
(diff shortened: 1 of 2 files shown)


williankl (changes requested):
one quick fix

williankl (commented):
I will verify if we have another way of exposing this without this file shenanigans, give me some hours pls 🙏🏾

williankl (approved):

DevSrSouza (approved):

caiqueslp (approved):

PR #150, opened 2026-04-16
feat: TopBar leading icon as slot with IconButton circular

caiqueslp:
## Summary

- Convert `TopBarAction` from `enum class` to `sealed class`, adding `Custom(icon: LemonadeIcons)` variant to support any icon as the TopBar leading navigation icon
- Replace manual `Box+Icon` rendering in `CoreTopBarActionContent` with `LemonadeUi.IconButton` using `Circular` shape (matching Figma design)
- Map `filled=true` to `Subtle` type, `filled=false` to `Ghost` type

### Breaking Changes

- `TopBarAction` is now a `sealed class` — `when` expressions need `is` keyword (e.g. `is TopBarAction.Back`)
- `TopBarAction.entries` no longer exists — use explicit list instead

## Test plan

- [ ] Verify TopBar Back icon renders as circular IconButton
- [ ] Verify TopBar Close icon renders as circular IconButton
- [ ] Verify `TopBarAction.Custom(icon)` renders the provided icon correctly
- [ ] Verify press state animation works (ripple/indication from IconButton)
- [ ] Verify `filled=true` shows subtle background, `filled=false` is transparent
- [ ] Verify collapsed/expanded states still work correctly
- [ ] Verify accessibility (contentDescription "Back"/"Close" for screen readers)

### DIFF
--- a/kmp/ui/src/mobileMain/kotlin/com/teya/lemonade/TopBar.mobile.kt
+++ b/kmp/ui/src/mobileMain/kotlin/com/teya/lemonade/TopBar.mobile.kt
@@ -806,11 +806,6 @@ private fun CoreTopBarActionContent(
         is TopBarAction.Close -> LemonadeIcons.Times
         is TopBarAction.Custom -> action.icon
     }
-    val contentDescription = when (action) {
-        is TopBarAction.Back -> "Back"
-        is TopBarAction.Close -> "Close"
-        is TopBarAction.Custom -> null
-    }
     val type = if (navigationAction.filled) {
         LemonadeButtonType.Subtle
     } else {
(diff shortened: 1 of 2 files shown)


kmp/ui/src/mobileMain/kotlin/com/teya/lemonade/TopBar.mobile.kt
```diff
@@ -803,34 +800,34 @@ private fun CoreTopBarActionContent(
...
+    val action = navigationAction.navigationAction
+    val icon = when (action) {
+        is TopBarAction.Back -> LemonadeIcons.ArrowLeft
+        is TopBarAction.Close -> LemonadeIcons.Times
+        is TopBarAction.Custom -> action.icon
+    }
+    val contentDescription = when (action) {
+        is TopBarAction.Back -> "Back"
+        is TopBarAction.Close -> "Close"
+        is TopBarAction.Custom -> null
```
williankl: "Back" and "Close" is not localized 👀 
maybe try to find a way to pass that responsibility to the call-site - if can't find a way to do so, maybe just leave it as null until we find a more solid solution, no?

williankl (changes requested):

williankl (approved):

PR #153, opened 2026-04-16
fix: prevent SegmentedControl indicator animation replay on recomposition

caiqueslp:
## Summary

- Fix SegmentedControl indicator replaying its entrance animation when the component scrolls out of view and back into view
- Root cause: `animateDpAsState` starts from `0.dp` on recomposition (since `remember` maps are cleared), then animates to the measured position with a bouncy spring
- Replace `animateDpAsState` with `Animatable` that `snapTo` on first measurement and `animateTo` on subsequent user-driven changes

## Test plan

- [ ] Open the SegmentedControl display screen on Android
- [ ] Scroll the SegmentedControl off screen and back — indicator should appear instantly without bouncy animation
- [ ] Tap different tabs — spring animation should still work normally
- [ ] Press and hold non-selected tabs — stretch animation should still work
- [ ] Verify on HomeDisplay (Light/Dark, Standard/Expressive segmented controls)

### DIFF
--- a/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/SegmentedControl.kt
+++ b/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/SegmentedControl.kt
@@ -2,7 +2,8 @@ package com.teya.lemonade
 
 import androidx.compose.animation.animateColorAsState
 import androidx.compose.animation.core.Spring
-import androidx.compose.animation.core.animateDpAsState
+import androidx.compose.animation.core.Animatable
+import androidx.compose.animation.core.VectorConverter
 import androidx.compose.animation.core.spring
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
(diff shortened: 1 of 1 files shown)


williankl (dismissed):

williankl (dismissed):

matheuslutero (approved):

kmp/ui/src/commonMain/kotlin/com/teya/lemonade/SegmentedControl.kt
```diff
@@ -185,19 +186,20 @@ internal fun CoreSegmentedControl(
         targetOffset = baseOffset
     }
 
+    val hasInitialMeasurement = remember { booleanArrayOf(false) }
+    val animationSpec = if (hasInitialMeasurement[0]) IndicatorSpringSpec else snap()
+
+    SideEffect {
+        if (hasMeasurements) hasInitialMeasurement[0] = true
+    }
```
DevSrSouza, line 194: Why does this `booleanArrayOf` is required?

caiqueslp: `booleanArrayOf` is used instead of `mutableStateOf` because this flag doesn't need to trigger recomposition on its own — what triggers recomposition is the change in `targetWidth`/`targetOffset` when measurements arrive via `onGloballyPositioned`.

If we used `mutableStateOf`, Compose would track a snapshot state read, and when `SideEffect` flips it to `true`, it would force an extra unnecessary recomposition. With `booleanArrayOf`, it's just a simple mutable wrapper that Compose doesn't track — the next natural recomposition (when the user interacts) will read the updated value and use `spring()` instead of `snap()`.

In short: avoids an extra no-op recomposition cycle.

DevSrSouza (approved):

PR #169, opened 2026-04-22
fix(ListItem): vertically center trailing slot + chevron (Android + iOS)

bruno-vsiqueira:
## Summary

- The middle `Row` (Android) / inner `HStack` (iOS) that wraps the content column plus the trailing slot was defaulting to top alignment. When a `ListItem` had both a `supportText` AND a navigation chevron / trailing slot, the chevron stuck to the title line instead of centering against the full content.
- **Android** (`kmp/ui/src/commonMain/kotlin/com/teya/lemonade/ListItem.kt`): added `verticalAlignment = Alignment.CenterVertically` to the wrapping `Row`.
- **iOS** (`swiftui/Sources/Lemonade/Components/LemonadeListItem.swift`): changed the inner `HStack` alignment from `.top` → `.center`. The outer `HStack` is left on `.top` so leading slots continue to top-align.
- **Fix B (included):** dropped the extra `.alpha(0.5)` / `.opacity(0.5)` layered over the already-semi-transparent `contentTertiary` tint on the enabled-state chevron — net ~20% opacity washed it out vs. Figma. Disabled-state dimming is preserved.
- **Follow-up (Felipe's iOS commit + Android mirror):** exposed `trailingAlignment` (iOS) / `trailingVerticalAlignment` (Android) on `ListItem` and `ActionListItem`, defaulting to `.center` / `Alignment.CenterVertically`. `ResourceListItem` pins it to `.top` internally so its value + addon column keeps top-aligning with the leading symbol. Sample apps on both platforms now include a dedicated `ActionListItem - Trailing Alignment` card.

Discovered while implementing MAC-5422 Card Controls in `teya-business-app` — cancel-card `ActionListItem(label, supportText, showNavigationIndicator = true)` reproduced the bug on both platforms.

### Bug fix — Notifications row (label + supportText + navigationIndicator)

Before: chevron stuck to title line and washed out. After: chevron vertically centered across label + support text, full tint opacity.

#### Android

| Before | After |
| --- | --- |
|  |  |

#### iOS

| Before | After |
| --- | --- |
|  |  |

### New `trailingAlignment` / `trailingVerticalAlignment` showcase

Top / Center / Bottom demonstrated with multi-line `supportText` + `Tag` trailing, matching between platforms.

| Android | iOS |
| --- | --- |
|  |  |

## Test plan

- [x] Android: Pixel 9 Pro emulator (API 36) — Notifications row chevron centered and darker; Trailing Alignment card renders Top / Center / Bottom correctly
- [x] iOS: iPhone 17 Pro Max simulator (iOS 26.2) — Notifications row chevron centered and darker; Trailing Alignment card renders .top / .center / .bottom correctly
- [ ] Verify disabled state still dims correctly on both platforms (`Disabled States` card in each showcase)

### DIFF
--- a/kmp/composeApp/src/commonMain/kotlin/com/teya/lemonade/ActionListItemDisplay.kt
+++ b/kmp/composeApp/src/commonMain/kotlin/com/teya/lemonade/ActionListItemDisplay.kt
@@ -10,6 +10,7 @@ import androidx.compose.foundation.layout.statusBarsPadding
 import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import com.teya.lemonade.core.LemonadeAssetSize
 import com.teya.lemonade.core.LemonadeBadgeSize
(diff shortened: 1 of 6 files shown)


hrafaelveloso (dismissed):

felipeemarcon (approved):
LGTM

williankl (approved):

PR #179, opened 2026-05-01
feat: add labelIsSecondary to ActionListItem (Android + iOS)

raoufteya:
Adds `labelIsSecondary` to `ActionListItem` on both platforms for the first time.

When `true`, the label renders small/grey (secondary style) and `supportText` renders as the primary text — flipping the visual hierarchy. Useful for patterns like "Last used / address" where the category label is subordinate to the value.

**Changes:**
- **Android (KMP):** new `labelIsSecondary` param on `ActionListItem`; renders label with `bodySmallRegular`/`contentSecondary` and supportText with `bodyMediumMedium`/primary color
- **iOS (SwiftUI):** same `labelIsSecondary` param wired through all `ActionListItem` overloads and `LemonadeListItem`
- **Demo:** new "ActionListItem - Label Is Secondary" card added to both Android and iOS sample apps

| Android | iOS |
|---------|-----|
| action list item - android | action list item - ios |

### DIFF
--- a/kmp/composeApp/src/commonMain/kotlin/com/teya/lemonade/ActionListItemDisplay.kt
+++ b/kmp/composeApp/src/commonMain/kotlin/com/teya/lemonade/ActionListItemDisplay.kt
@@ -79,6 +79,45 @@ internal fun ActionListItemDisplay() {
             )
         }
 
+        // ActionListItem - Label Is Secondary
+        LemonadeUi.Card(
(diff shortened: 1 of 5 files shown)


williankl (changes requested):
This really breaks the API - not sure if a good approach. I recommend first checking with if this is really what we want to do here in this UI - if yes, I recommend using the base `LemonadeUi.ListItem` with a `contentSlot` instead.

PR #181, opened 2026-05-01
feat(swiftui, kmp): add verticalAlignment to ContentListItem

raoufteya:
## Summary
- Add `verticalAlignment` parameter to `ContentListItem` horizontal layout on **SwiftUI** and **KMP**
- Fix KMP layout bug where long value text crushed the label column to one character wide (added `weight(1f)` to value row)
- Add long text samples to both SwiftUI and KMP sample apps to demonstrate multi-line wrapping behavior

## Screenshots

| Platform | Before | After |
|----------|--------|-------|
| **Android** | android-before-1 | android-after-1 |
| **iOS** | ios-before-1 | ios-after-1 |

## Test plan
- [ ] Verify short label/value pairs still render correctly (no regression)
- [ ] Verify long text wraps properly with default center alignment
- [ ] Verify `verticalAlignment: .top` / `Alignment.Top` aligns label to top when value wraps
- [ ] Verify value text is always right-aligned in horizontal layout
- [ ] Test on both iOS simulator and Android emulator

### DIFF
--- a/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/ContentListItem.kt
+++ b/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/ContentListItem.kt
@@ -49,7 +49,6 @@ import com.teya.lemonade.core.SymbolContainerVoice
  * @param contentSlot - Optional slot for additional content. In vertical layout, this also
  *   switches the value typography to bodyXLargeSemiBold.
  * @param verticalAlignment - Vertical alignment for horizontal layout (default [Alignment.CenterVertically]).
- * @param valueTextAlign - Text alignment for the value in horizontal layout (default [TextAlign.End]).
  */
 @Composable
 public fun LemonadeUi.ContentListItem(
@@ -59,7 +58,6 @@ public fun LemonadeUi.ContentListItem(
     modifier: Modifier = Modifier,
     showDivider: Boolean = false,
     verticalAlignment: Alignment.Vertical = Alignment.CenterVertically,
-    valueTextAlign: TextAlign = TextAlign.End,
     leadingSlot: (@Composable RowScope.() -> Unit)? = null,
     trailingSlot: (@Composable RowScope.() -> Unit)? = null,
     contentSlot: (@Composable ColumnScope.() -> Unit)? = null,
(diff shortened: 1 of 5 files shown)


kmp/ui/src/commonMain/kotlin/com/teya/lemonade/ContentListItem.kt
```diff
@@ -56,6 +58,8 @@ public fun LemonadeUi.ContentListItem(
     layout: LemonadeContentListItemLayout = LemonadeContentListItemLayout.Horizontal,
     modifier: Modifier = Modifier,
     showDivider: Boolean = false,
+    verticalAlignment: Alignment.Vertical = Alignment.CenterVertically,
+    valueTextAlign: TextAlign = TextAlign.End,
```
williankl: Not sure about this API 🤔 
@felipeemarcon should we let the value alignment be changed? in this case we better just use a slot instead, and let the user decide what to up in the end, no?

williankl (changes requested):
let's wait for @felipeemarcon feedback here

kmp/ui/src/commonMain/kotlin/com/teya/lemonade/ContentListItem.kt
williankl: Just to be more precise, I like the `verticalAlignment` api, but not sure about the `valueTextAlign` 😬

williankl: @raoufteya I checked the Figma and I didnt see a case where it is not right aligned - can you point out to me where it is used please?

raoufteya: Delivery To is right aligned 
image

caiqueslp (dismissed):

kmp/ui/src/commonMain/kotlin/com/teya/lemonade/ContentListItem.kt
felipeemarcon: The value in the horizontal variant should always be right-aligned. @raoufteya @williankl.

If the purpose of `valueTextAlign` is only to allow right-aligned, it should not be necessary.

Regarding `verticalAlignment`, it seems sensible to cover some possible use cases. I'm okay with this one.

raoufteya: done

williankl (approved):
Thanks for the adjustments 🚀

Issue #196, opened 2026-05-24
perf: ListItem has a high per-row render cost (~5ms/frame floor) that causes scroll jank in long lists


caiqueslp:
## Summary

`LemonadeUi.ListItem` (the `CoreListItem` implementation in `kmp/ui/src/commonMain/kotlin/com/teya/lemonade/ListItem.kt`) has a **high irreducible per-row render cost**. Profiling a long scrolling list shows that even a **label-only** `ListItem` (no leading/trailing slots) costs roughly **~1.8 ms measure/layout + ~2.3 ms draw + ~1.0 ms recompose per frame** during a fling. On a 60 Hz device that's a ~16.6 ms budget, so composing/measuring/drawing only a handful of newly-revealed rows per frame already blows the budget → scroll jank.

Because `ListItem` backs essentially every scrolling list across the Teya apps, this is a cross-app performance ceiling, not a single-screen issue.

## Where this was found

- App: `teya-business-app`, Sales tab → **Transactions list** (a `LazyColumn` of `ListItem` rows).
- Device: **Pixel 10a, Android 16**, debug build.
- Lemonade version: **`com.teya.foundation:lemonade-ui:0.15.5`**, Compose `1.10.2`, Kotlin `2.3.20` (strong skipping on).
- Method: Perfetto system traces (`adb shell perfetto`, `android.surfaceflinger.frametimeline` + ftrace atrace for the app), analyzed with `trace_processor`. Per-frame phase cost = average duration of the app main-thread slices `AndroidOwner:measureAndLayout`, `Record View#draw()`, `Compose:recompose` during the scroll.

## Evidence (A/B)

We measured the same Transactions scroll with three row variants. The decisive one is **full row vs. label-only row** (label-only = `LemonadeUi.ListItem(label, supportText, isLoading, showDivider, onListItemClick)` with **no** `leadingSlot` / `trailingSlot`):

| Per-frame cost | Full row (icon + amount + status tag) | **Label-only `ListItem` (DS floor)** |
|---|---|---|
| `AndroidOwner:measureAndLayout` | ~3.1 ms | **~1.79 ms** |
| `Record View#draw()` | ~2.5 ms | **~2.34 ms** |
| `Compose:recompose` | ~1.76 ms | **~1.0 ms** |

Key reading:
[...]

PR #197, opened 2026-05-24
perf(list-item): cut per-row render cost (draw floor, gating, layout)

caiqueslp:
## What

Reduces the per-row render cost of `LemonadeUi.ListItem` (`CoreListItem`) that causes scroll jank in long lists, per #196.

The profiling in #196 attributes the per-row floor to three things, all addressed here:

| # | Source (from #196) | Change |
|---|---|---|
| 1 | Per-row rounded **clip layer + animated background** (~2.3 ms draw) | Draw-phase rounded fill highlight; no clip layer |
| 2 | Always-on `animateColorAsState` + interaction collectors | Gated behind `onListItemClick != null` |
| 3 | Layout nesting (~6 Row + ~5 Column / row) | SafeArea collapse + skip empty trailing `Row` |

## How

**1. Clickable rows — draw-phase fill instead of clip + ripple (the big one).**
The old path kept a persistent `Modifier.clip(radius500)` graphics layer on every row purely to bound the Material ripple, plus an `animateColorAsState` background drawn each frame. The clip layer is the dominant draw cost during a fling. Replaced with a `drawWithCache`/`drawOutline` highlight that reads the animated colour **at draw time**: while scrolling the colour is transparent, so the row draws nothing and allocates no layer. Press/hover feedback is now an animated rounded fill (no ripple).

This intentionally changes Android press feedback from a ripple to a fill — chosen to match the **existing iOS `ListItemButtonStyle`** (`.background(isPressed ? color : .clear).clipShape(...)`), so the two platforms are now consistent.

**2. Non-clickable rows — pixel-identical gating.**
A row without `onListItemClick` has no `clickable`, so it can never be pressed/hovered; its background always resolved to `interactionBackground.copy(alpha = opacity0)` and `opacity0 = 0f`. Removing the clip + background + `animateColorAsState` + collectors for those rows is therefore provably pixel-identical.

**3. Layout trims (no visual change).**
- `SafeArea`: collapse the redundant outer `Column` in the common no-divider case; drop two no-op `.background(Color.Transparent)`.
- Skip the trailing `Row` when there is no trailing slot and no navigation indicator (it was an empty 0-width `Row` composed/measured on every row).

The deeper measure win (a single custom `Layout`, M3-style) is intentionally **not** in this PR — it needs visual coverage we don't have yet and is the natural follow-up.

## Verification

- ✅ `compileDebugKotlinAndroid`, `ktlintCheck`, `apiCheck` (public API unchanged), `detektAndroidDebug`
- ✅ Press highlight confirmed on a **Pixel 10a (Android 16)** via the sample app
- ✅ Cross-checked the approach against **Material 3 `ListItem`** (static `Surface` vs. `InteractiveListItem`; keeps clip only for the ripple) and the **Lemonade SwiftUI `ListItem`** (gates the highlight behind a click handler; fill, no ripple)
- ⏳ On-device scroll-jank A/B (Perfetto) on the real `teya-business-app` Transactions list — to be confirmed by the issue author

## Behavior change for reviewers / design

Press/hover feedback on **clickable** rows changes from a Material ripple to an animated rounded fill. This matches iOS. Non-clickable rows, disabled rows, and all static appearance are unchanged.

Refs #196

## API Dump

**Verdict:** NO_CHANGES

No public API changes — the change is an internal refactor of `CoreListItem`/`SafeArea` rendering. Confirmed both ways: the committed `api/*.api` / `*.klib.api` baseline is unchanged vs `main`, and a fresh `./gradlew apiDump` regenerates an identical baseline (so `apiCheck` passes).

### DIFF
--- a/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/ListItem.kt
+++ b/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/ListItem.kt
@@ -581,17 +581,20 @@ private fun CoreListItem(
     SafeArea(modifier = modifier, showDivider = showDivider) {
         Row(
             verticalAlignment = Alignment.CenterVertically,
-            modifier = if (onListItemClick != null) {
-                Modifier.interactiveBackground(
-                    interactionSource = interactionSource,
-                    voice = voice,
-                    enabled = enabled,
-                    role = role,
-                    onClick = onListItemClick,
-                )
-            } else {
-                Modifier
-            }.defaultMinSize(minHeight = LocalSizes.current.size1200)
+            modifier = Modifier
+                .then(
+                    other = if (onListItemClick != null) {
+                        Modifier.interactiveBackground(
+                            interactionSource = interactionSource,
+                            voice = voice,
+                            enabled = enabled,
+                            role = role,
+                            onClick = onListItemClick,
+                        )
+                    } else {
+                        Modifier
+                    },
+                ).defaultMinSize(minHeight = LocalSizes.current.size1200)
                 .padding(
                     horizontal = LocalSpaces.current.spacing300,
                     vertical = LocalSpaces.current.spacing300,
@@ -735,7 +738,10 @@ private fun SafeArea(
         return
     }
 
-    Column(modifier = modifier) {
+    Column(
+        verticalArrangement = Arrangement.Center,
+        modifier = modifier,
+    ) {
         Column(
             verticalArrangement = Arrangement.Center,
             modifier = Modifier.padding(all = LocalSpaces.current.spacing100),
--- a/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/ListItem.kt
+++ b/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/ListItem.kt
@@ -1,7 +1,6 @@
 package com.teya.lemonade
 
 import androidx.compose.animation.animateColorAsState
-import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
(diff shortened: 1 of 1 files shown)


kmp/ui/src/commonMain/kotlin/com/teya/lemonade/ListItem.kt
```diff
@@ -590,59 +573,119 @@ private fun CoreListItem(
...
+ * ripple); press feedback is the animated fill, mirroring the iOS `ListItemButtonStyle`.
+ */
+@Composable
+private fun Modifier.interactiveBackground(
+    interactionSource: MutableInteractionSource,
+    voice: LemonadeListItemVoice,
+    enabled: Boolean,
+    role: Role?,
+    onClick: () -> Unit,
+): Modifier {
```
williankl, line 689: I like this 👍🏾

kmp/ui/src/commonMain/kotlin/com/teya/lemonade/ListItem.kt
```diff
@@ -519,37 +519,20 @@ private fun CoreListItem(
...
+                Modifier.interactiveBackground(
+                    interactionSource = interactionSource,
+                    voice = voice,
+                    enabled = enabled,
+                    role = role,
+                    onClick = onListItemClick,
+                )
+            } else {
+                Modifier
+            }.defaultMinSize(minHeight = LocalSizes.current.size1200)
```
williankl: instead of this if-else, use the `.then(...)` modifier pls

matheuslutero (approved):

PR #229, opened 2026-06-12
Match tile's min height on android and ios

GustavoASantos:
# Summary
Tile's min height was applied after the padding on android, while on ios it was applied before the padding, leading to a taller min height on ios. I removed the min height before the padding on ios, leaving the one in tile shape which is after the padding

## Screenshot

 
IMG_0028

## API Dump

**Verdict:** 

### DIFF
--- a/swiftui/Sources/Lemonade/Components/LemonadeTile.swift
+++ b/swiftui/Sources/Lemonade/Components/LemonadeTile.swift
@@ -459,7 +459,7 @@ private struct LemonadeTileSlotView<LeadingContent: View, TopAccessory: View>: V
                 leadingSlot()
                 tileTextContent
             }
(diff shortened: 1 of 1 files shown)


swiftui/Sources/Lemonade/Components/LemonadeTile.swift
```diff
@@ -459,7 +459,7 @@ private struct LemonadeTileSlotView<LeadingContent: View, TopAccessory: View>: V
                 leadingSlot()
                 tileTextContent
             }
-            .frame(minWidth: minWidth, maxWidth: .infinity, minHeight: minHeightHorizontal, alignment: .leading)
```
GustavoASantos, line 462: The min height is still applied in tileShape, after the padding

caiqueslp (approved):

PR #231, opened 2026-06-12
chore(ui): make generated LemonadeRes class internal (publicResClass = false)

DevSrSouza:
# Summary

Sets `publicResClass = false` in `kmp/ui/build.gradle.kts` so the Compose Resources–generated `LemonadeRes` accessor class (and its `Drawable*_commonMainKt` / `Font0_commonMainKt` facades) is `internal` instead of `public`, and regenerates the `ui` API baseline accordingly.

## Why

The Compose Resources plugin partitions the generated `gen_*` drawable accessors across auto-numbered `Drawable0..N_commonMainKt` facade files. Adding icons grows the set and **reshuffles** accessors between facades — and on the JVM the facade class is part of each top-level accessor's binary signature. So every icon batch reads as an ABI break to the **API Stability Review** gate, even though nothing a consumer can name actually changed. This is exactly what blocked #228 (icons sliding from `Drawable4` → `Drawable5`).

Consumers never use the raw `LemonadeRes.drawable.gen_*` accessors. They go through the typed public APIs — `LemonadeIcons`, `LemonadeCountryFlags`, `LemonadeBrandLogos` — whose extensions (`LemonadeIconsExtension.kt` etc.) resolve the `gen_*` accessors **internally, within the `ui` module**. Marking `LemonadeRes` internal keeps those extensions compiling (same-module access) while removing the volatile generated surface from the public ABI. After this lands, adding icons never touches a public symbol again, so icon PRs stop tripping the gate.

### Verified no one depends on the removed surface
- All **562** `LemonadeRes` references in this repo are inside the `ui` module itself (only `ui` has a `compose.resources` block). No cross-module use; `composeApp` has zero.
- **teya-business-app** and **teya-terminal-application** (both on lemonade `0.18.0`): **zero** references to `LemonadeRes`, `gen_*`, or `Drawable*_commonMainKt`. They consume icons exclusively via `LemonadeIcons.*` (1003 / 346 usages respectively).

## API Dump

**Verdict:** BREAKING

- Every one of the ~2375 removed declarations is a **Compose Resources–generated** symbol — `LemonadeRes` and its nested `$drawable` / `$font` / `$string` / `$array` / `$plurals` objects, the `Drawable0..5_commonMainKt` / `Font0_commonMainKt` facades, the `getGen_*` accessors, `ActualResourceCollectorsKt`, and the associated `$stable` fields. None are hand-written, and none are referenced by name in consumer source.
- No hand-authored public declaration is removed or changed. The diff is **removals only** (`+0` additions across all three `ui` api files); the typed icon/flag/logo APIs in `core`/`ui` are untouched.
- Source compatibility is fully preserved for the supported API: `LemonadeIcons.X`, `LemonadeIcons.drawableResource`, and `LemonadeUi.Icon(icon = …)` are unchanged. A consumer recompiling against this release sees no difference.

**Genuine breaking changes:**
- This is a deliberate, one-time removal of the generated `LemonadeRes` surface from the public ABI. It is technically ABI-breaking for any consumer that linked pre-compiled bytecode referencing `LemonadeRes.drawable.gen_*` / `Drawable*_commonMainKt` directly — but no known consumer does (see verification above), and that surface was never the intended icon API.
- Per repo policy, a BREAKING verdict needs sign-off on the head commit from one of **@caiqueslp**, **@williankl**, or **@DevSrSouza**. Re-request review after any new push — approval is matched to the commit SHA.


williankl (approved):
go drinking

PR #250, opened 2026-06-21
feat(BottomTabBar): add optional badge on tab items

caiqueslp:
## What

`BottomTabBarItem` gains an optional `badge: String?`. When set, a small `LemonadeUi.Tag` renders over the icon's top-trailing corner via Material3 `BadgedBox` — so a tab can flag a **"Soon" / "New"** state without shifting the icon or label. `null` keeps the current icon-only layout (no behavior change for existing call sites — the param is defaulted).

Driven by the app needing a "Soon" badge on the **Teya AI** tab while the feature is behind a flag.

## Changes

- `BottomTabBarItem.badge: String? = null` (+ KDoc).
- `BottomTabBarItemContent` wraps the icon in `BadgedBox`, rendering `LemonadeUi.Tag(label = badge, voice = .Neutral)` when present.
- Binary-compat `.api` dumps regenerated (`apiDump`): android / desktop / klib.
- Demo (`BottomTabBarDisplay`) shows the badge on the "Teya AI" item.

## Design notes / for review

- **Voice:** used `TagVoice.Neutral` (gray pill). The app's reference design (web) shows a **lime/brand** "Soon" pill — there's no brand `TagVoice` today, so the exact color needs a design/token call. Happy to switch to whichever voice (or a new brand voice) you prefer.
- **Positioning:** `BadgedBox` default top-end placement; a text Tag is wider than a dot badge, so on very narrow slots it can sit close to the edge. Open to a tighter/size-capped badge variant if preferred.

## Verify

- `:expressive:compileDebugKotlinAndroid` ✅
- `:expressive:apiDump` ✅ (committed)

### DIFF
--- a/kmp/composeApp/src/commonMain/kotlin/com/teya/lemonade/BottomTabBarDisplay.kt
+++ b/kmp/composeApp/src/commonMain/kotlin/com/teya/lemonade/BottomTabBarDisplay.kt
@@ -59,6 +59,7 @@ private fun LazyListScope.defaultBottomTabBarSection() {
                     label = "Teya AI",
                     icon = LemonadeIcons.SparklesSoft,
                     selectedIcon = LemonadeIcons.SparklesSoftSolid,
+                    badge = "Soon",
                 ),
             ),
             selectedIndex = 0,
--- a/kmp/expressive/api/android/expressive.api
+++ b/kmp/expressive/api/android/expressive.api
@@ -10,13 +10,17 @@ public final class com/teya/lemonade/BottomSheet_androidKt {
 
 public final class com/teya/lemonade/BottomTabBarItem {
 	public static final field $stable I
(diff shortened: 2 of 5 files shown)


caiqueslp:
## API Dump (updated after binary-compat fix)

**Verdict: `ADDITIONS_ONLY`** (confirmed via `.claude/skills/binary-compatibility/scripts/bcv-check.sh`) — CI auto-passes, no maintainer gate needed.

`badge: String?` was **appended** to the `BottomTabBarItem` data class (never inserted), and the pre-badge `<init>` / `copy` / `copy$default` symbols are restored as `@Deprecated(HIDDEN)` shims. Those retained symbols dump as `synthetic` with identical descriptors, so the classifier counts the `-`/`+` pairs as additive rather than removals. Already-compiled consumers keep linking whether they call the old constructor, `copy(...)`, or `copy()`.

No property renames/removals, no abstract interface members, no enum changes.

williankl (approved):

PR #280, opened 2026-07-13
fix(date-picker): align the iOS weekday header with the day grid's first weekday

Filmaluco:
# Summary

Fixes a visual misalignment in the SwiftUI `DatePicker` / `DateRangePicker` month grid, reported by a UK merchant through the Teya app (Zendesk #1106576): July 13 2026 — a Monday — rendered under the "S" (Sunday) header column.

**Root cause:** the weekday header rendered the caller's Sunday-first `weekdayAbbreviations` array verbatim, while `CalendarUtils.generateMonthDays` places day cells starting at `Calendar.current.firstWeekday`. Those two conventions only coincide on Sunday-first regions (e.g. US). On Monday-first regions (UK and most of Europe) every date landed one column left of its label. Date *selection* was always correct — the bug was purely the header/grid alignment.

**Fix:** rotate the header labels by the same calendar's `firstWeekday` before rendering (`CalendarUtils.orderedWeekdayAbbreviations`), so header and grid share one source of truth. The documented `weekdayAbbreviations` contract (Sunday→Saturday) is unchanged, so no caller needs to change. Both `DatePicker` and `DateRangePicker` are fixed — they share `CoreDatePickerView`.

The `InlineCalendar` strip is unaffected (each cell carries its own weekday label, and its `weekdayLabels` helper already honours `firstWeekday`).

## Repro

On a device with Region = United Kingdom (or any Monday-first region), open any screen using `LemonadeUi.DatePicker`. The grid lays out Monday-first while the header stays Sunday-first — all dates shift one column. US-region devices/simulators mask the bug.

## Screenshot

Merchant screenshot (Teya app, "Filter by period"): July 1 (a Wednesday) under "T", July 13 (Monday) under "S". After the fix the header reads M T W T F S S on Monday-first devices, matching the grid.

## Tests

Added `CalendarUtilsTests`:
- label rotation for Sunday-first / Monday-first / Saturday-first calendars,
- the reported case — July 13 2026 must sit under "Mon" on a Monday-first calendar,
- an exhaustive invariant: for every `firstWeekday` 1–7, every one of the 42 grid cells' weekday matches its column label.

`swift test`: 20 tests, 0 failures.

## API Dump

No public API changes (Swift-only change; no `kmp/<module>/api` files touched).

### DIFF
--- a/swiftui/Sources/Lemonade/Components/Calendar/CalendarUtils.swift
+++ b/swiftui/Sources/Lemonade/Components/Calendar/CalendarUtils.swift
@@ -146,6 +146,24 @@ enum CalendarUtils {
         return "\(monthFormatter(month)) \(year)"
     }
 
+    /// Rotates 7 Sunday-first weekday labels so index 0 is the calendar's `firstWeekday`.
+    ///
+    /// A month grid from ``generateMonthDays(year:month:calendar:)`` places its first column
+    /// on `calendar.firstWeekday`, so a header row must be rotated with this before rendering
+    /// or its labels drift from the day cells on any non-Sunday-first calendar.
+    ///
+    /// - Parameters:
+    ///   - abbreviations: Exactly 7 labels ordered Sunday through Saturday. Anything past the
+    ///     first 7 is dropped; fewer than 7 are returned unchanged.
+    ///   - calendar: The calendar whose `firstWeekday` decides the rotation.
+    /// - Returns: The labels reordered to start at the calendar's first weekday.
+    static func orderedWeekdayAbbreviations(_ abbreviations: [String], calendar: Calendar) -> [String] {
+        let labels = Array(abbreviations.prefix(7))
+        guard labels.count == 7 else { return labels }
+        let first = calendar.firstWeekday - 1
+        return (0..<7).map { labels[(first + $0) % 7] }
+    }
+
     /// Returns an array of 7 weekday labels starting from the calendar's first weekday.
     ///
     /// - Parameters:
--- a/swiftui/Sources/Lemonade/Components/Calendar/CalendarUtils.swift
+++ b/swiftui/Sources/Lemonade/Components/Calendar/CalendarUtils.swift
@@ -178,7 +178,6 @@ enum CalendarUtils {
         case .short:
             symbols = calendar.shortWeekdaySymbols
         }
-        let first = calendar.firstWeekday - 1 // 0-indexed
-        return (0..<7).map { symbols[(first + $0) % 7] }
+        return orderedWeekdayAbbreviations(symbols, calendar: calendar)
     }
 }
--- a/swiftui/Sources/Lemonade/Components/LemonadeDatePicker.swift
+++ b/swiftui/Sources/Lemonade/Components/LemonadeDatePicker.swift
@@ -92,7 +92,9 @@ public extension LemonadeUi {
     ///   - state: Configuration state holding the selected date, min/max bounds.
     ///     Observe ``LemonadeDatePickerState/selectedDate`` to react to user selections.
     ///   - monthFormatter: Formatter that returns the month name for a given month number (1-12).
-    ///   - weekdayAbbreviations: Exactly 7 items representing Sunday through Saturday.
+    ///   - weekdayAbbreviations: Exactly 7 items representing Sunday through Saturday. The header
+    ///     renders them starting from the current calendar's first weekday, so the labels stay
+    ///     aligned with the day grid on every device region.
     ///   - today: The date treated as "today" — drives both the "today" indicator and the pager's
     ///     initial month. Defaults to `Date()`. Callers can pass a different value (e.g. the
     ///     currently-selected day) to have the pager open on that day's month; the day will also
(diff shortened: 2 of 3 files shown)


swiftui/Sources/Lemonade/Components/Calendar/CalendarUtils.swift
```diff
@@ -146,6 +146,24 @@ enum CalendarUtils {
...
+    /// - Parameters:
+    ///   - abbreviations: Exactly 7 labels ordered Sunday through Saturday. Anything past the
+    ///     first 7 is dropped; fewer than 7 are returned unchanged.
+    ///   - calendar: The calendar whose `firstWeekday` decides the rotation.
+    /// - Returns: The labels reordered to start at the calendar's first weekday.
+    static func orderedWeekdayAbbreviations(_ abbreviations: [String], calendar: Calendar) -> [String] {
+        let labels = Array(abbreviations.prefix(7))
+        guard labels.count == 7 else { return labels }
+        let first = calendar.firstWeekday - 1
+        return (0..<7).map { labels[(first + $0) % 7] }
```
matheuslutero, line 164: Same rotation math as `weekdayLabels` below, it could just delegate to this now

swiftui/Sources/Lemonade/Components/LemonadeDatePicker.swift
```diff
@@ -335,7 +339,10 @@ private struct CoreDatePickerView: View {
             Spacer().frame(height: LemonadeTheme.spaces.spacing200)
             
             HStack(spacing: 0) {
-                ForEach(Array(weekdayAbbreviations.prefix(7).enumerated()), id: \.offset) { _, day in
+                ForEach(
+                    Array(CalendarUtils.orderedWeekdayAbbreviations(weekdayAbbreviations, calendar: calendar).enumerated()),
```
matheuslutero, line 343: same on compose? `DatePicker.kt` header renders the sunday-first abbreviations verbatim while the grid starts at `firstDayOfWeek`, same drift for monday-first callers. worth a follow up?

GustavoASantos (dismissed):

swiftui/Sources/Lemonade/Components/LemonadeDatePicker.swift
Filmaluco: Does not apply to to Compose

swiftui/Sources/Lemonade/Components/Calendar/CalendarUtils.swift
```diff
@@ -146,6 +146,24 @@ enum CalendarUtils {
...
+    ///
+    /// - Parameters:
+    ///   - abbreviations: Exactly 7 labels ordered Sunday through Saturday. Anything past the
+    ///     first 7 is dropped; fewer than 7 are returned unchanged.
+    ///   - calendar: The calendar whose `firstWeekday` decides the rotation.
+    /// - Returns: The labels reordered to start at the calendar's first weekday.
+    static func orderedWeekdayAbbreviations(_ abbreviations: [String], calendar: Calendar) -> [String] {
+        let labels = Array(abbreviations.prefix(7))
+        guard labels.count == 7 else { return labels }
+        let first = calendar.firstWeekday - 1
```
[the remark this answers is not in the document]
Filmaluco: Checked this empirically: Foundation clamps `firstWeekday` on read — assigning `0`, `9`, or `-3` to a `Calendar` all read back as `1`:

```swift
var c = Calendar(identifier: .gregorian)
c.firstWeekday = 0   // reads back 1
c.firstWeekday = 9   // reads back 1
c.firstWeekday = -3  // reads back 1
```

So `firstWeekday` is always `1...7` at the point of use and the negative-remainder index can't happen through the `Calendar` API (this component only ever passes `Calendar.current` anyway — same pattern `weekdayLabels` has always used). Skipping the normalization to avoid guarding an unreachable state.

swiftui/Sources/Lemonade/Components/Calendar/CalendarUtils.swift
Filmaluco: Good call — done in 72b3e71c, `weekdayLabels` now delegates to `orderedWeekdayAbbreviations`.

swiftui/Sources/Lemonade/Components/LemonadeDatePicker.swift
Filmaluco: Checked this while diagnosing the original bug: Compose is self-consistent **by default** — `firstDayOfWeek` defaults to a hardcoded `DayOfWeek.SUNDAY` (`DatePicker.kt:307/363/437/457`) and never consults device locale, so the Sunday-first header matches the grid out of the box. The drift only appears when a caller passes `firstDayOfWeek = MONDAY` (etc.) without reordering the abbreviations themselves — the component doesn't rotate the header from the param. So: no user-facing bug today, but the API lets callers recreate this exact misalignment. Agreed it's worth a follow-up to rotate the Compose header from `firstDayOfWeek` (mirror of this fix); keeping it out of this PR since it's KMP public-API-adjacent and needs the BCV pass.

matheuslutero (approved):

PR #281, opened 2026-07-13
fix(SymbolContainer): align 2XL to 72dp / radius 20dp with Figma

williams-oliveira:
## What
Fixes the **2XL (XXLarge)** `SymbolContainer` size, which had drifted from the Figma spec:
- container **80dp → 72dp** (`size2000` → `size1800`)
- Rounded corner radius **24dp → 20dp** (`radius600` → `radius500`)

Applied on both platforms: **KMP** (`SymbolContainer.kt`) and **SwiftUI** (`LemonadeSymbolContainer.swift`). `contentSize`, icon size and text style are unchanged.

## Why
Figma defines 2XL as 72dp / 20dp radius; the implementation had drifted to 80dp / 24dp. This realigns the component with the source of truth.

## Impact
Affects every consumer of `SymbolContainer` size **XXLarge** (~25 usages in the Teya app: side menu, terminals, settlement accounts/preferences, risk visibility, card front, etc.) — each shrinks by 8dp with slightly tighter corners. Intended visual change.

## Validation
Rendered the real component on an Android emulator, before vs after:

| | Container | Radius |
|---|---|---|
| Before (`main`) | 80dp | 24dp |
| After (this branch) | 72dp | 20dp |

Fill image and icon both clip cleanly to the new radius. _(Before/after screenshot attached below.)_
symbolcontainer_2xl_before_after

## API
No public API change — the `XXLarge` enum entry is unchanged, only its internal value. `apiCheck`: **NO_CHANGES**.

### DIFF
--- a/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/SymbolContainer.kt
+++ b/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/SymbolContainer.kt
@@ -351,7 +351,7 @@ private fun SymbolContainerShape.resolveShape(size: SymbolContainerSize): Shape 
             SymbolContainerSize.Medium -> LocalShapes.current.radius300
             SymbolContainerSize.Large -> LocalShapes.current.radius400
             SymbolContainerSize.XLarge -> LocalShapes.current.radius500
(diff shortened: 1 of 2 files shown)


felipeemarcon (approved):
LGTM

PR #283, opened 2026-07-15
feat(bottom-sheet): add gesturesEnabled flag

Norbi444:
## What

Adds a `gesturesEnabled` parameter to `LemonadeUi.BottomSheet` (the common overload and the Android `hideNavigationBar` overload).

When `gesturesEnabled = false`:
- swipe/drag gestures are disabled via Material3's `sheetGesturesEnabled`;
- the drag handle is hidden (it's a non-functional affordance without gestures).

Scrim-tap and back-press dismissal stay controlled independently through `properties` (`dismissOnClickOutside` / `dismissOnBackPress`), so the caller composes exactly the lock they want.

## Why

The earlier `dismissible` flag gated `SheetValue.Hidden` through `confirmValueChange`. That same predicate also backs the programmatic `sheetState.hide()` (via `AnchoredDraggableState.anchoredDrag`), so a non-dismissible sheet could never be closed by the caller setting `expanded = false` — the Close button did nothing. Disabling only the gesture (rather than vetoing the Hidden state) leaves programmatic close working.

## Sample

The "Non-dismissible" catalog example now composes the locked behaviour explicitly:

```kotlin
LemonadeUi.BottomSheet(
    expanded = showNonDismissibleSheet,
    onDismissRequest = { showNonDismissibleSheet = false },
    gesturesEnabled = false,
    properties = LemonadeBottomSheetProperties(
        dismissOnBackPress = false,
        dismissOnClickOutside = false,
    ),
) { … }
```

Verified on device (A920Pro, Android 8.1.0): scrim tap ✅ ignored, back press ✅ ignored, swipe ✅ blocked (no handle), Close button ✅ dismisses.

## API Dump

**Verdict: `ADDITIONS_ONLY`** — CI auto-passes.

- The new `gesturesEnabled: Boolean` (a `Boolean`, appended before `background`) changes the public function's descriptor, adding a new overload. The previously released signatures are preserved as `@Deprecated(level = HIDDEN)` overloads, so the old symbols remain and the classifier sees only additions.
- The `-` line on the prior public signature is just the `synthetic` flag being added (descriptor unchanged) — that's the `HIDDEN` shim, not a removal.
- The `getLambda$<hash>$expressive` singletons re-hashed because a `@Composable` gained a parameter — internal, name-mangled, never consumer-visible; counted as additive.

No genuinely breaking changes.


williankl (approved):
LGTM

PR #284, opened 2026-07-15
Add ToastAnchor to position a toast above a bottom action button

eduard-sergeev:
# Summary
Toasts currently always anchor to the very bottom edge, so on screens with a fixed bottom action button, the toast overlaps it.                                                                                
                                                                                                                                                                                                                 
  Adds a ToastAnchor (ToastAnchor / LemonadeToastAnchor on Swift) with two cases:                                                                                                                                
  - Bottom (default, unchanged) — flush with the bottom edge.                                                                                                                                                    
  - AboveBottomActionButton — 112dp/pt bottom padding, clearing a standard bottom action button.                                                                                                                 
                                                                                                                                                                                                                 
  Settable via LemonadeToastState.show(...) (Kotlin) and LemonadeToastManager.show(...) (Swift).                                                                                                                 
                                                                                                                                                                                                                 
  Also fixes an iOS hit-testing bug: the toast's tap/drag gesture area extended into its bottom padding via contentShape(Rectangle()), so anything visible underneath the toast (e.g. an action button) was
  unclickable while a toast was shown. The anchor's bottom padding is now excluded from the interactive region.

## Figma component
<!-- Provide figma link (if any) to the component, it makes it easy for reviewers to find it -->

## Screenshot
<!-- Provide at least one image of your change if applicable -->
<!-- <img src="drawing.jpg" alt="drawing" width="200"/> -->
<!-- You can define the width/height to resize the image if needed -->

### Teya Business App
| android | ios |
|--------|--------|
| <img width="1080" height="2400" alt="Screenshot_1784127449" src="https://github.com/user-attachments/assets/88529292-5906-45e6-b185-da5f2daaf16b" /> | <img width="1206" height="2622" alt="Simulator Screenshot - iPhone 17 Pro - 2026-07-15 at 18 10 24" src="https://github.com/user-attachments/assets/9f5cc970-da56-4c46-88bf-8fd8caa69c9e" /> | 

### Demo
| Default | Above action button |
|--------|--------|
| <img width="1080" height="2400" alt="Screenshot_1784543932" src="https://github.com/user-attachments/assets/2f855504-8179-46ac-896f-9ede4c2f8c08" /> | <img width="1080" height="2400" alt="Screenshot_1784544009" src="https://github.com/user-attachments/assets/0c0e2fee-c2c6-4a7f-b46f-ca5b930e81a4" /> | 


## API Dump
<!--
Required when any kmp/<module>/api/*.api or *.klib.api file changed.
Run `.claude/skills/binary-compatibility/scripts/bcv-check.sh --ci` to get the verdict.
If nothing in api/ changed, write "No public API changes" and delete the rest.
-->

**Verdict:** <!-- NO_CHANGES / ADDITIONS_ONLY / BREAKING -->

<!--
- Interface addition — the interface is local-only, no external implementers.
- Enum entry addition — config enum; the component owns the `when`.
- `@Deprecated(HIDDEN)` overload — the `-` line is only the `synthetic` flag,
  the JVM descriptor is unchanged, so already-compiled consumers keep linking.
-->

<!--
List each one, who needs to approve (@caiqueslp / @williankl / @DevSrSouza),
and why the break is acceptable. Leave empty if the verdict is not BREAKING.
-->

### DIFF
--- a/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/Toast.kt
+++ b/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/Toast.kt
@@ -193,6 +193,34 @@ public class LemonadeToastState {
         )
     }
 
+    @Deprecated(
+        message = "Use show() with an anchor parameter to position the toast.",
+        replaceWith = ReplaceWith(
+            expression = "show(label, voice, icon, duration, dismissible, actionLabel, onAction, ToastAnchor.Bottom)",
+        ),
+        level = DeprecationLevel.HIDDEN,
+    )
+    public fun show(
+        label: String,
+        voice: ToastVoice = ToastVoice.Neutral,
+        icon: LemonadeIcons? = null,
+        duration: ToastDuration = ToastDuration.Short,
+        dismissible: Boolean = true,
+        actionLabel: String? = null,
+        onAction: (() -> Unit)? = null,
+    ) {
+        show(
+            label = label,
+            voice = voice,
+            icon = icon,
+            duration = duration,
+            dismissible = dismissible,
+            actionLabel = actionLabel,
+            onAction = onAction,
+            anchor = ToastAnchor.Bottom,
+        )
+    }
+
     /** Programmatically dismiss the current toast. */
     public fun dismiss() {
         currentToast = null
--- a/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/Toast.kt
+++ b/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/Toast.kt
@@ -13,7 +13,10 @@ import androidx.compose.foundation.clickable
 import androidx.compose.foundation.gestures.detectVerticalDragGestures
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.calculateEndPadding
+import androidx.compose.foundation.layout.calculateStartPadding
 import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.navigationBarsPadding
 import androidx.compose.foundation.layout.offset
@@ -36,6 +39,7 @@ import androidx.compose.ui.hapticfeedback.HapticFeedbackType
 import androidx.compose.ui.input.pointer.pointerInput
 import androidx.compose.ui.layout.onSizeChanged
 import androidx.compose.ui.platform.LocalHapticFeedback
+import androidx.compose.ui.platform.LocalLayoutDirection
 import androidx.compose.ui.semantics.Role
 import androidx.compose.ui.semantics.role
 import androidx.compose.ui.semantics.semantics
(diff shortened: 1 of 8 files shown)


DevSrSouza (dismissed):

kmp/ui/src/commonMain/kotlin/com/teya/lemonade/Toast.kt
```diff
@@ -352,6 +395,7 @@ private fun CoreToast(
 }
 
 private const val DRAG_DISMISS_THRESHOLD_DP = 25
+private const val ABOVE_BOTTOM_ACTION_BUTTON_PADDING_DP = 112
```
felipeemarcon: Would not a fixed value for this be rather sensitive? If, for some reason, the height of the fixed footer changes due to more content or something, this size will not work as expected, right?

Is there a way to guarantee it gets the correct needed size?

williankl:
Please, wait a bit before merging this 🙏🏾

kmp/ui/src/commonMain/kotlin/com/teya/lemonade/Toast.kt
williankl: we cannot guarantee it AFAIK - I'm thinking on a solution here - but do we really need this feature at all?

kmp/ui/src/commonMain/kotlin/com/teya/lemonade/Toast.kt
```diff
@@ -94,6 +105,7 @@ internal data class ToastData(
     val id: Int,
     val actionLabel: String?,
     val onAction: (() -> Unit)?,
+    val anchor: ToastAnchor,
 )
```
williankl: instead of a `ToastAnchor`, use the Jetpack Compose `Alignment` interface, depending on the interface we can place it anywhere, apply that in the alignment space for the content.
add another parameter `paddingValues: PaddingValues` to address padding.

does this solve it for you?

williankl (changes requested):
Left a suggestion to address this

kmp/ui/src/commonMain/kotlin/com/teya/lemonade/Toast.kt
eduard-sergeev: that should work, but that means that you can set any padding -> the padding could be inconsistent between different app screens

felipeemarcon: @eduard-sergeev I'm "okay" with possible different padding values. Two reasons:

1. Depending on the screen on the app, it will need a different value
2. I believe the app, should, somehow, handle this value through a helper or something. The component only receive the value using a param.

Makes sense?

eduard-sergeev: @felipeemarcon , how about ios?
`PaddingValues` doesn't exist on iOS. It's a Jetpack Compose type. The Swift side (LemonadeToastManager.show(...)) has no equivalent — today ToastAnchor/LemonadeToastAnchor are two intentionally mirrored enums, same case names, same meaning. Switching Kotlin to PaddingValues breaks that symmetry outright; Swift would need a completely different shape (CGFloat? EdgeInsets?)

williankl: can we not use `EdgeInsets` in this case?

williankl (approved):
Looks good

PR #290, opened 2026-07-22
fix(ui): load asset drawables without blocking the main thread

williankl:
# Summary

A consumer app hit an ANR whose trace ends in `CountryFlagKt.CoreCountryFlag` → compose-resources `painterResource`. On non-web targets, compose-resources resolves `painterResource(DrawableResource)` through `rememberResourceState` (`ResourceState.blocking.kt`), which is literally `remember { mutableStateOf(runBlocking { load() }) }` — the main thread opens the vector XML from the APK, reads it, and parses it during composition. On a contended cold start, the cumulative first-load cost of every asset on screen crossed the ~5s input-dispatch timeout.

This PR removes that main-thread block for all three components that load resources internally (`Icon`, `CountryFlag`, `BrandLogo`):

- New internal `rememberAssetPainter(resource)` (`AssetPainter.kt`): decodes the vector on `Dispatchers.Default` via `getDrawableResourceBytes` + `decodeToImageVector`, renders a transparent painter for the (first-load-only) in-flight frames, and caches decoded `ImageVector`s for the process lifetime so every later composition resolves synchronously — same steady-state behavior as today, minus the block.
- `LocalInspectionMode` falls back to the synchronous `painterResource`, keeping previews and screenshot tests pixel-identical.
- Each call site is a one-line swap; layout is unchanged (the existing `requiredSize` modifiers reserve the slot, so the asset late-paints without reflow).

Known gap (possible follow-up): slot APIs where consumers build painters themselves via `painterResource` (e.g. `Chip`'s `leadingPainter`) still block on the consumer side; a public async helper would be ADDITIONS_ONLY.

## Benchmark (before vs after)

Measured with the `composeApp` sample — debug APKs built from `main` (`f61fbd3e`, **before**) and this branch (**after**) — on an Android 16 (API 36) arm64 emulator. Stress surface: the **CountryFlag display screen** (~250 flags at `XXLarge` in a lazy grid). Each run: fresh process → open the flags screen → two scroll flings over a ~5s window; frame stats from `dumpsys gfxinfo` (reset right before opening the screen).

**Flags screen — cold I/O (first run after a fresh APK install), the ANR scenario:**

| Metric | Before (blocking) | After (async) |
|---|---|---|
| Frames rendered in the window | 33 | **148** |
| Janky frames | 21 (**63.6 %**) | 5 (**3.4 %**) |
| Frame time p50 / p90 | 48 / 150 ms | 17 / 20 ms |
| Frame time p95 / p99 | **400 / 650 ms** | **24 / 81 ms** |

Before, every newly composed flag runs blocking I/O + XML parsing on the main thread: individual frames reached 650 ms *on an emulator backed by a fast host SSD* — on low-end hardware with contended flash at cold start, that same work is what crossed the ~5 s input-dispatch timeout. After, the worst frame is 81 ms and the UI produced 4.5× more frames in the same wall-clock window.

**Flags screen — warm OS page cache (second fresh-process run):** before 4.8 % janky / p99 77 ms vs after 5.4 % / p99 65 ms — statistically identical; the process-wide `ImageVector` cache keeps steady-state behavior unchanged.

**Cold app start (`am start -W`, 3 runs):** before 999/805/792 ms vs after 858/776/802 ms — unchanged within noise.

**Rendering verification:** screenshots on the after build confirm all three components render correctly — full flags grid, 279-icon grid (tint applied), and all brand logos including the largest asset (SEPA).

*Caveats: debug builds, emulator, 1–2 runs per cell — directional comparison; real low-end hardware amplifies the "before" numbers.*

## Figma component

N/A — no visual/design change.

## Screenshot

N/A — rendering is unchanged; only the loading mechanism moved off the main thread (see benchmark section for on-device verification).

## API Dump

No public API changes.

**Verdict:** NO_CHANGES (`bcv-check.sh --ci`: "public API is identical")

### DIFF
--- a/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/AssetPainter.kt
+++ b/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/AssetPainter.kt
@@ -1,0 +1,67 @@
+package com.teya.lemonade
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.produceState
+import androidx.compose.runtime.remember
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.painter.ColorPainter
+import androidx.compose.ui.graphics.painter.Painter
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.graphics.vector.rememberVectorPainter
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.platform.LocalInspectionMode
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import org.jetbrains.compose.resources.DrawableResource
+import org.jetbrains.compose.resources.ExperimentalResourceApi
+import org.jetbrains.compose.resources.decodeToImageVector
+import org.jetbrains.compose.resources.getDrawableResourceBytes
+import org.jetbrains.compose.resources.painterResource
+import org.jetbrains.compose.resources.rememberResourceEnvironment
+
+private val assetVectorCache = mutableMapOf<DrawableResource, ImageVector>()
+
+/**
+ * Loads a vector [DrawableResource] without blocking the main thread.
+ *
+ * compose-resources' [painterResource] resolves synchronously on non-web targets
+ * (`runBlocking` during composition), which can ANR on contended cold starts. This
+ * helper decodes on a background dispatcher instead, rendering a transparent painter
+ * until the first decode of each resource completes. Decoded vectors are cached for
+ * the process lifetime, so every later composition resolves synchronously.
+ *
+ * [LocalInspectionMode] keeps previews and screenshot tests on the synchronous path.
+ */
+@OptIn(ExperimentalResourceApi::class)
+@Composable
+internal fun rememberAssetPainter(resource: DrawableResource): Painter {
+    if (LocalInspectionMode.current) {
+        return painterResource(resource = resource)
+    }
+    val density = LocalDensity.current
+    val environment = rememberResourceEnvironment()
+    val imageVector by produceState(
+        initialValue = assetVectorCache[resource],
+        key1 = resource,
+    ) {
+        if (value == null) {
+            value = withContext(context = Dispatchers.Default) {
+                getDrawableResourceBytes(
+                    environment = environment,
+                    resource = resource,
+                ).decodeToImageVector(density = density)
+            }.also { decodedVector ->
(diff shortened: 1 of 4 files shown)


williankl:
## Benchmark: before vs after (sample app on Android emulator)

Measured with the `composeApp` sample — debug APKs built from `main` (`f61fbd3e`, **before**) and this branch (**after**) — on an Android 16 (API 36) arm64 emulator. The stress surface is the **CountryFlag display screen**, which composes all ~250 flags at `XXLarge` in a lazy grid. Each run: fresh process → open the flags screen → two scroll flings over a ~5 s window; frame stats from `dumpsys gfxinfo` (reset right before opening the screen).

### Flags screen — cold I/O (first run after a fresh APK install)

| Metric | Before (blocking) | After (async) |
|---|---|---|
| Frames rendered in the window | 33 | **148** |
| Janky frames | 21 (**63.6 %**) | 5 (**3.4 %**) |
| Frame time p50 | 48 ms | 17 ms |
| Frame time p90 | 150 ms | 20 ms |
| Frame time p95 | **400 ms** | **24 ms** |
| Frame time p99 | **650 ms** | **81 ms** |

Before, every newly composed flag runs blocking I/O + XML parsing on the main thread: individual frames reached 650 ms *on an emulator backed by a fast host SSD*. On low-end hardware with contended flash at cold start, that same work is what crossed the ~5 s input-dispatch timeout and produced the reported ANR. After, the worst frame is 81 ms and the UI produced 4.5× more frames in the same wall-clock window — input dispatch is never starved. The scripted gestures also traveled visibly further through the grid on the async build because scrolling never stalled.

### Flags screen — warm OS page cache (second fresh-process run)

| Metric | Before | After |
|---|---|---|
| Janky frames | 4.8 % | 5.4 % |
| Frame time p99 | 77 ms | 65 ms |

Statistically identical — the process-wide `ImageVector` cache keeps steady-state behavior unchanged, as intended. This also shows how deceptive the blocking path is: it only ANRs when I/O is cold/contended, which is exactly the cold-start scenario consumers hit.

### Cold app start (`am start -W` TotalTime, 3 runs)

Before: 999 / 805 / 792 ms · After: 858 / 776 / 802 ms — unchanged within noise (the home screen composes few assets).

### Rendering verification

Screenshots on the after build confirm all three components render correctly with the async loader: the full flags grid, the 279-icon grid (tint applied), and all brand logos including the largest asset (SEPA).

*Caveats: debug builds, emulator, 1–2 runs per cell — treat as a directional comparison; real low-end hardware amplifies the "before" numbers.*

caiqueslp (approved):

DevSrSouza (approved):

PR #300, opened 2026-07-29
fix(swiftui): follow Apple's per-style Dynamic Type curves

caiqueslp:
Follows the text-scaling proposal in #299. This is step 1 from its order of work: the one change that touches every string the SwiftUI package draws.

## The problem

Every Lemonade text style built its font through `relativeTo: .body`. Apple ships one Dynamic Type curve per text style precisely because they are not meant to match — text that starts big is already legible, so it is designed to grow less. Resolving everything through `.body` handed a 72pt display style the growth curve meant for 17pt body copy.

Measured in a consumer app at roughly AX4: body copy sat at 2.25x and a 48pt balance at 2.06x. Near-identical multipliers, which is the fingerprint of a single curve applied to everything. The balance ran to three lines and pushed the rest of the screen out of view.

## The change

`LemonadeTextStyle` now carries the Apple text style whose curve it follows, and each style names one.

Switching on `fontSize` alone could not work: `displayXSmall` and `headingSmall` are both 24pt, `displaySmall` and `headingMedium` both 28pt, and each pair wants a different curve. So it is a stored property, not a derived one.

| Lemonade tier | size | curve |
|---|---|---|
| `display*` | 24–72 | `.largeTitle` |
| `headingXLarge` | 40 | `.title` |
| `headingLarge`, `headingMedium` | 32, 28 | `.title2` |
| `headingSmall` | 24 | `.title3` |
| `headingXSmall`, `headingXXSmall` | 18, 16 | `.headline` |
| `bodyXLarge*`, `bodyLarge*`, `bodyMedium*` | 20, 18, 16 | `.body` |
| `bodySmall*` | 14 | `.footnote` |
| `bodyXSmall*` | 12 | `.caption` |

It also threads the curve through the four places that built fonts outside `LemonadeTextStyle.font`:

- `LemonadeText.resolveFont()` (String overload) — this one dropped `relativeTo:` entirely
- `LemonadeText.resolveFont()` (AttributedString overload)
- the markdown inline attributes, so a bold run scales with the text around it
- `LemonadeBadge`, which sets its font from a size token rather than a text style

## Measured

Run through `UIFontMetrics` on the simulator, iOS 26.4. `old` is what shipped, `new` is this branch.

| style | base | AX3 old | AX3 new | AX5 old | AX5 new | AX5 change |
|---|---|---|---|---|---|---|
| `display3XLarge` | 72 | 157 (2.18x) | 107 (1.49x) | 203 (2.82x) | 123 (1.71x) | **−39%** |
| `displayLarge` | 48 | 104.7 (2.18x) | 71.3 (1.49x) | 135.3 (2.82x) | 82 (1.71x) | **−39%** |
| `headingLarge` | 32 | 69.7 (2.18x) | 59.3 (1.85x) | 90.3 (2.82x) | 75.3 (2.35x) | −17% |
| `bodyMediumRegular` | 16 | 35 (2.19x) | 35 (2.19x) | 45 (2.81x) | 45 (2.81x) | 0% |
| `bodySmallRegular` | 14 | 30.7 (2.19x) | 31 (2.21x) | 39.3 (2.81x) | 40.3 (2.88x) | +3% |
| `bodyXSmallRegular` | 12 | 26.3 (2.19x) | 29.3 (2.44x) | 33.7 (2.81x) | 38.3 (3.19x) | **+14%** |

Small text growing more is the intended direction, not a side effect. `.footnote` and `.caption` are steeper curves than `.body`, and small text is what most needs the room at accessibility sizes. It does mean dense components have more text to hold, which is what the layout floors in the rest of #299 are for.

**At the default content size nothing moves.** I screenshotted the sample app on `main` and on this branch at `content_size large` and diffed them: the only differing pixels were the status bar clock (the two runs were two minutes apart). Every rendered glyph is identical.

At AX5 in the sample app, `display3XLarge` went from wrapping onto two lines with two and a half styles visible on screen, to fitting on one line with all eight styles visible.

## The mapping lands where Android already sits

Measured on a Pixel 10a (Android 16) after the fact, to check these tier boundaries against a platform that already solves this. The ceilings differ — Android stops at 2.0x, iOS AX5 goes past 2.8x — so absolute growth does not compare. The ratio between tiers does.

| | `display3XLarge` | `bodyMedium` | display as % of body |
|---|---|---|---|
| Android at `font_scale 2.0` | 1.04x | 1.75x | **59%** |
| iOS at AX5, this branch | 1.71x | 2.81x | **61%** |
| iOS at AX5, `main` | 2.82x | 2.81x | 100% |

59% against 61%, reached independently — the mapping here came from Apple's size tables without looking at Android.

That is the argument I would lead with if this gets pushback along the lines of "design wants our sizes kept". Nothing moves at the default content size, which the pixel diff above already shows. What changes is that the same design token used to grow 12% on Android and 182% on iOS. The platforms disagreed with each other, and the 100% row is the outlier that needed explaining. This is convergence on behaviour Android already ships, not a taste call.

Android needs no counterpart change, and the reason is structural. Apple keys the curve on a named style, so it is a parameter separate from the size and you can ask for the wrong one — which we did, for every style. Android keys on the sp value itself, so the size is the key and there is nothing to ask for. `LemonadeTypographyExtensions.kt:47-48` already emits `.sp` for both `fontSize` and `lineHeight`.

One caveat, for the record rather than for action: `minSdk` is 23 and Android's converter arrived in 14, so pre-14 devices fall back to linear scaling and do have the problem this fixes. There is no API to address it — the conversion tables are not on the device.

## Tests

Ten new assertions in `LemonadeDynamicTypeCurveTests`. Six pin the mapping and run today under `swift test` (35 tests pass). Four measure real growth through `UIFontMetrics` and assert display grows more slowly than body at AX5, that nothing scales at default, and that no style ends up smaller than body copy at the same setting — rule 6 of the contract.

One thing worth flagging: those four need UIKit, so they compile out on macOS, and the generated Xcode project has no test target — `swiftui_ci.yml` only builds, it never runs tests. So the four measured assertions do not execute anywhere right now, and neither do the four test files already in the repo. Adding a `bundle.unit-test` target to `project.yml` and a `xcodebuild test` step to CI is a small follow-up that would turn all of this into a real safety net. Happy to do it separately rather than bundle CI changes into a behaviour PR.

## Compatibility

The new parameter is last and defaults to `.body`, so existing call sites keep compiling and any `LemonadeTextStyle` built outside the design system behaves exactly as before. Same for the new `toLemonadeMarkdown(relativeTo:)` argument.

No KMP modules touched, so no API baseline movement. Android already gets the right behaviour for free — `LemonadeTypographyExtensions.kt:47` emits plain `.sp`, which goes through Android 14's non-linear converter.

## Reviewing this

The tier boundaries are a judgement call and the part worth arguing about. `headingXSmall` and `headingXXSmall` land on `.headline`, which is body-sized semibold, so they grow at the body rate rather than a title rate. That seemed right for 16–18pt text, but it is the row I would push back on first.

## Validated against the real app

Wired into `teya-business-app` as a local Swift package (`apps/teyaApp/ios/project.yml`, `path:` instead of `url:`/`from:`) and ran `Teya Dev` on a simulator at AX5 — the same screen and setting that started this investigation.

Measuring glyph bands on the home screen, normalised to logical points (the two captures had different pixel scales, so raw pixels do not compare):

| | before | after | |
|---|---|---|---|
| balance digit height | 73.0pt | 50.3pt | **−31%** |
| `interest` (body copy) | 28.4pt | 29.0pt | unchanged |

Body copy holding still is what makes the comparison trustworthy: both captures are at the same accessibility setting, so the 31% drop is the curve change and not a different slider position.

In practice the screen used to end at "earned this month" with the tab bar overlapping it. It now fits the Sell / Pay buttons, the Notifications header, and the first notification card.

Two things this does **not** fix, both consumer-side rather than design-system:

- The balance still wraps to two lines, and the `£` drops onto the second line by itself. That is composition — the currency symbol and the amount are separate views in a horizontal arrangement that breaks at a bad point. 31% smaller helps but does not fit a seven-digit amount on one line. The fix belongs on that screen (one string, or a `minimumScaleFactor` on the group).
- The app's bottom tab bar labels stay frozen at roughly 12pt while everything around them is at ~2.8x. That is the opposite bug — text that should grow and does not. There is no `LemonadeBottomTabBar` in the SwiftUI package, so this bar is the app's own and wants its own ticket.


felipeemarcon (dismissed):
LGTM

PR #306, opened 2026-07-30
build(swiftui): move signing settings to a gitignored Local.xcconfig

felipeemarcon:
## Problem

`swiftui/Lemonade.xcodeproj` is generated by XcodeGen and gitignored. A signing
team set through Xcode's **Signing & Capabilities** editor lives only in that
generated project, so the next regeneration wipes it and device builds fail with:

> error: Signing for "LemonadeSampleApp" requires a development team.

The workaround is passing `DEVELOPMENT_TEAM=<id>` on every `xcodebuild` call.

## Approach

Signing settings move to `swiftui/Local.xcconfig` — gitignored, referenced from
`project.yml` as the base configuration — matching how the Teya business app
handles this.

`scripts/generate-ios-project.sh` creates the file from
`Local.xcconfig.template` and fills in your Team ID when it finds exactly one
signing certificate in your keychain, so the usual case needs no manual step:

```bash
cd swiftui && ./scripts/generate-ios-project.sh
# 📝 Created Local.xcconfig with DEVELOPMENT_TEAM = XXXXXXXXXX
```

When the keychain is ambiguous or empty it stays blank and prints how to set it,
per the README:

```bash
security find-certificate -c "Apple Development" -p | openssl x509 -noout -subject
# take the OU field → swiftui/Local.xcconfig
```

An existing `Local.xcconfig` is never overwritten.

## Why not commit the Team ID

This repository is **public**. A pinned team ID isn't a credential leak — team
IDs ship in the `application-identifier` entitlement of every distributed IPA,
and signing still needs certificates and private keys that aren't here. But it
makes every fork sign against an organisation its contributors don't belong to,
failing with a misleading error. Gitignoring the file also means it can't be
committed by accident.

An earlier revision of this branch read the team from an exported environment
variable. A file scoped to the checkout is better than shell state that leaks
into every other Xcode project on the machine.

## CI and publishing

Both workflows now generate through the script instead of calling `xcodegen
generate`, because XcodeGen **refuses to generate when a referenced config file
is absent**. Runners have no signing certificate, so `DEVELOPMENT_TEAM` is left
empty — harmless, because neither pipeline consults it:

- `swiftui_ci.yml` builds both schemes for `generic/platform=iOS Simulator` with
  `CODE_SIGNING_ALLOWED=NO`.
- `swiftui_release.yml` runs `scripts/build_xcframework.sh`, which archives for
  device and simulator, also with `CODE_SIGNING_ALLOWED=NO`.

## Verification

Ran the real commands rather than reasoning about them:

| Scenario | Result |
|---|---|
| Clean checkout → generate script | created `Local.xcconfig`, auto-filled the Team ID |
| Device build, no CLI override, no env var | `** BUILD SUCCEEDED **` (iPhone 17 Pro Max) |
| Empty team → CI framework build | `** BUILD SUCCEEDED **` |
| Empty team → CI sample-app build | `** BUILD SUCCEEDED **` |
| Empty team → release archive, `generic/platform=iOS` | `** ARCHIVE SUCCEEDED **` |
| No signing certificate at all (CI runner) | script completes, project generated |
| Existing `Local.xcconfig` | left untouched |
| `git check-ignore` on `Local.xcconfig` | ignored |
| `git grep` for any Team ID in tracked files | absent |

The no-certificate case caught a real bug: under `set -euo pipefail`, the
detection helper returned non-zero when it found nothing, which aborted the
script before generating the project — it would have broken both workflows.
Fixed and covered by the run above.

## API Dump

No public API changes.


swiftui/scripts/generate-ios-project.sh
```diff
@@ -0,0 +1,60 @@
...
+
+    # Ambiguous (several teams) or absent: leave it blank rather than guess wrong.
+    if [ "$(printf '%s' "$teams" | grep -c . || true)" -eq 1 ]; then
+        printf '%s' "$teams"
+    fi
+
+    # Must succeed even when nothing was found: under `set -e` a non-zero return
+    # here would abort the caller's `TEAM_ID=$(detect_team_id)` assignment.
+    return 0
+}
```
williankl, line 41: Turns out `openssl x509` only reads the first cert you pipe into it, so the ambiguity check never fires — with certs from two teams in the keychain this silently picks whichever comes first. I tried it with two fake certs to confirm.

Counting the certs before parsing fixes it. Side effect: two certs of the *same* team (e.g. after a renewal) now also leave it blank, but I think that's fine — blank is the safe default here, and the script already prints what to do.

```suggestion
detect_team_id() {
    local pems
    pems=$(security find-certificate -a -c "Apple Development" -p 2>/dev/null) || return 0

    # openssl x509 reads only the first PEM block, so several certificates
    # cannot be told apart after the pipe — refuse to guess unless there is
    # exactly one.
    if [ "$(printf '%s\n' "$pems" | grep -c 'BEGIN CERTIFICATE')" -ne 1 ]; then
        return 0
    fi

    # `|| return 0`: under `set -e` a non-zero pipeline here would abort the
    # caller's `TEAM_ID=$(detect_team_id)` assignment.
    printf '%s\n' "$pems" \
        | openssl x509 -noout -subject 2>/dev/null \
        | sed -n 's/.*OU *= *\([A-Z0-9]*\).*/\1/p' \
        || return 0
}
```

Same first-cert-only thing applies to the manual `security find-certificate` one-liner in the README btw — not blocking.

williankl (dismissed):
Went through this with a security hat on since it touches signing, CI and the gitignore of a public repo. Couldn't find anything to worry about: no team id was ever committed (I checked the force-pushed revision too), the gitignore entry works, the script never touches the network, and PR builds have no secrets in scope. Nice 👏

Found one small bug in the team detection — left a suggestion inline.

Also, maybe drop a quick note in the SwiftUI channel that plain `xcodegen generate` stops working after this merges — the error it gives doesn't mention the script.

Great PR, loved the verification table. Approving.

swiftui/scripts/generate-ios-project.sh
felipeemarcon: Good catch, and confirmed — applied in 19faed4.

I reproduced it before fixing, with two self-signed certs carrying different `OU`s and a stubbed `security`:

| keychain holds | before | after |
|---|---|---|
| no certs | blank | blank |
| one cert | `TEAMAAAA111` | `TEAMAAAA111` |
| certs from two teams | **`TEAMAAAA111`** — silently wrong | blank |
| same team twice (renewal) | `TEAMAAAA111` | blank |

Took your version as-is; I only extended the comment to spell out the same-team-twice case, so nobody "fixes" it back later thinking it's an oversight. Agreed blank is the right default — simulator builds are unaffected and the script already prints what to do.

Also added `-a` to the manual one-liner everywhere it appears, since it had the same first-cert-only problem: `README.md`, `Local.xcconfig.template`, and the script's own help output.

Noted on the SwiftUI channel heads-up about bare `xcodegen generate` — that error really doesn't hint at the script.

The new commit dismissed your approval, so this is blocked on a re-approve at 19faed4 when you have a moment. @williankl 🙏

williankl (approved):
vaicomdeus

PR #323, opened 2026-08-17
perf(list-item): cut the remaining per-row cost (indication node, wrapper collapse, measure policy, draw paths)

williankl:
Refs #196.

PR #197 removed the headline draw floor (persistent clip + animated background). This PR takes the levers that remained, plus the same class of cost in the row's usual neighbours (SymbolContainer leading icons, skeletons, the outlined SelectListItem, the theme `clickable`). Every public signature is untouched; the entire restructure is in private/internal code.

## What changed (one commit per lever, safest first within each area)

**ListItem**
- **Interaction machinery → `IndicationNodeFactory`** (`ListItemHighlight.kt`, new): the press/hover fill used to be a `@Composable` modifier factory running two interaction collectors + `animateColorAsState` per clickable row (3 coroutines, eager `Animatable`, recomposition on every press/hover flip, modifier chain rebuilt each recomposition). It is now a `Modifier.Node` indication handed to `clickable`: one collector in the node's scope, a fraction `Animatable` created lazily on first interaction, theme colour/shape read at draw time, and only `invalidateDraw()` ever fires — interaction never recomposes the row. Animation is `spring()` on a 0→1 fraction scaling the colour's alpha, trajectory-identical to the previous same-RGB colour fade.
- **Overload chain collapsed**: `ActionListItem`/`ResourceListItem`/string-`ListItem` now call the private `CoreListItem` directly (shared text stack extracted into an inline `ListItemTextContent`). ActionListItem's path drops from 7 restartable scopes per row to 3. The 8 `@Deprecated(HIDDEN)` shims are untouched (they already bind to live overloads; four vestigial `@Suppress("DEPRECATION")` removed).
- **SafeArea → modifiers**: the wrapper Column(s) around every row are gone; gutter padding and the divider are plain modifiers on the row itself. Divider rows draw the line in `drawBehind` with its height reserved as bottom padding, mirroring `CoreHorizontalDivider`'s solid-variant tokens and stroke maths exactly (−3 nodes, −2 composable frames per divider row).
- **Body `Row` + `BoxWithConstraints` → one `MeasurePolicy`**: `priority = Label` no longer pays a per-row SubcomposeLayout (composition during measure); both priorities are one two-child measure policy (content forced to the remainder for Trailing; content capped to preserve the `size2000` trailing floor for Label, trailing end-anchored). `placeRelative` throughout, so RTL mirrors as before. Side effect: `IntrinsicSize` on a Label-priority row used to **crash** (SubcomposeLayout throws on intrinsics) and now works. Rows with no trailing content skip the body wrapper entirely.
- Node math per row (no divider): label-only 4 → 2 layout nodes; label+icon 5 → 3; full row with divider 8 → 5; Label-priority full row also loses the per-row subcomposition.

**Neighbours**
- **SymbolContainer**: the container fill is drawn with the shape (`background(color, shape) + border` — the Chip idiom, and what SwiftUI's SymbolContainer already does) instead of clipping every symbol, so icons/text/fit-painters cost no clip layer. Fill-painters and `contentSlot` keep the clip (content genuinely overflows). The per-instance `CompositionLocalProvider` for platform dimensions is gone — dimensions are a slot parameter now.
- **Skeleton**: the shimmer allocated a fresh gradient `Brush` every frame and clipped an inner Box. Now one Box with `drawWithCache`: brush + corner radius cached per size, each frame pans the cached shader via canvas translate (identical gradient span), zero per-frame allocation, no clip.
- **Theme `clickable`** (`Modifier.kt`): dropped the redundant `composed {}` — the function is already `@Composable`, so it reads `LocalEffects` directly. Body-only change.
- **Outlined SelectListItem**: `clip + background + border` → `background(color, shape) + border` (no clip layer per row). Selection animation untouched. (A draw-phase selected-surface node replacing the two `animateColorAsState` was considered and deferred — the recompose-on-toggle is inherent to selection, and the clip removal was the measurable win.)

## Measurements

Debug builds, emulator SP-2 (Android 16, 720×1280), a local 400-row stress screen (not committed) of full transaction-style rows (leading `SymbolContainer`, support text, trailing text, chevron, divider, clickable), 12 scripted identical flings, atrace slice totals normalised per drawn frame — the same phase metric as the issue. Two alternating runs per side.

| Per-frame phase | before | after | Δ |
|---|---|---|---|
| **Trailing priority** — `Record View#draw()` | 0.84 ms | 0.69 ms | **−18%** |
| **Trailing** — `Compose:recompose` | 0.88 ms | 0.76 ms | **−14%** |
| **Trailing** — `measureAndLayout` | 1.71 ms | 1.71 ms | ~0% (text-measurement-dominated; node savings within emulator noise) |
| **Label priority** — `measureAndLayout` | 1.86 ms | 1.30 ms | **−30%** |
| **Label** — `Record View#draw()` | 0.75 ms | 0.62 ms | **−17%** |
| **Label** — `Compose:recompose` | 0.73 ms | 0.57 ms | **−21%** (recompose slice count 485 → 267, the measure-time subcompositions are gone; reproduced exactly across runs) |

Happy to re-run the numbers on a physical device or share the raw traces; @caiqueslp, your Perfetto A/B on the Transactions list would be the definitive check.

## Visual parity

Screenshot pixel-diffs (before vs after builds, status bar cropped):
- Stress rows, both priorities: identical except ≤4/255 anti-aliasing deltas on the symbol-circle edge (clip-layer AA vs shape-fill AA), ~0.2% of pixels.
- Press highlight (mid-long-press capture): identical bounds, identical changed-pixel count (157,102 both sides), identical fill colour.
- `ResourceListItemDisplay` (the Top-aligned trailing / two-level alignment-box case): identical except the same symbol-edge AA.
- `SkeletonDisplay`: ≤2/255.
- `SelectListItemDisplay`: Plain sections byte-identical; Outlined sections differ only in ≤6/255 corner AA plus scattered single pixels (max 51/255) along the selected item's curved border — same AA class.

## Behaviour notes (documented divergences, all edge-case)

- A caller forcing a ListItem taller than natural: the row now stretches (no divider) / the divider pins to the bottom (with divider) instead of the old Column centring. No such caller exists in the repo.
- The divider now renders in unbounded-width contexts (it previously collapsed to zero width via `fillMaxWidth` under an infinite max) — arguably a fix.
- Changing `voice` mid-press resets the highlight animation (indication factory inequality swaps the node) instead of retargeting its colour.
- In expressive themes, the outlined SelectListItem's ripple is no longer rounded by a clip — the same behaviour Chip already ships.
- `contentSlot` SymbolContainers still clip (SwiftUI does not) — kept conservative; relaxing it is a cheap follow-up.

## API Dump

**Verdict: `ADDITIONS_ONLY`** (verified with `bcv-check.sh --ci`; klib dumps unchanged).

The only baseline delta is one removed line in the android + desktop JVM dumps:

```
- public final fun getLambda$1280770520$ui_release ()Lkotlin/jvm/functions/Function3;
```

This is a `ComposableSingletons$ListItemKt` lambda getter — a hash-named, module-internal implementation detail of a composable body that no consumer can reference across modules; collapsing the overload chain removed the non-capturing lambda it backed. Not consumer-visible; the classifier counts it as additive. All 12 public `ListItemKt` overloads (4 live + 8 HIDDEN shims), `SymbolContainer`, `SelectListItem`, and the theme `clickable` keep their exact signatures.

### DIFF
--- a/kmp/ui/api/android/ui.api
+++ b/kmp/ui/api/android/ui.api
@@ -128,7 +128,6 @@ public final class com/teya/lemonade/ComposableSingletons$ListItemKt {
 	public final fun getLambda$-1242978266$ui_release ()Lkotlin/jvm/functions/Function3;
 	public final fun getLambda$-1808195552$ui_release ()Lkotlin/jvm/functions/Function3;
 	public final fun getLambda$-965205749$ui_release ()Lkotlin/jvm/functions/Function3;
-	public final fun getLambda$1280770520$ui_release ()Lkotlin/jvm/functions/Function3;
 	public final fun getLambda$787311865$ui_release ()Lkotlin/jvm/functions/Function3;
 }
 
--- a/kmp/ui/api/desktop/ui.api
+++ b/kmp/ui/api/desktop/ui.api
@@ -128,7 +128,6 @@ public final class com/teya/lemonade/ComposableSingletons$ListItemKt {
 	public final fun getLambda$-1242978266$ui ()Lkotlin/jvm/functions/Function3;
 	public final fun getLambda$-1808195552$ui ()Lkotlin/jvm/functions/Function3;
 	public final fun getLambda$-965205749$ui ()Lkotlin/jvm/functions/Function3;
-	public final fun getLambda$1280770520$ui ()Lkotlin/jvm/functions/Function3;
 	public final fun getLambda$787311865$ui ()Lkotlin/jvm/functions/Function3;
 }
 
--- a/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/ListItem.kt
+++ b/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/ListItem.kt
@@ -2,11 +2,8 @@
 
 package com.teya.lemonade
 
-import androidx.compose.animation.animateColorAsState
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.interaction.MutableInteractionSource
-import androidx.compose.foundation.interaction.collectIsHoveredAsState
-import androidx.compose.foundation.interaction.collectIsPressedAsState
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.BoxScope
@@ -21,14 +18,11 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.layout.widthIn
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.alpha
-import androidx.compose.ui.draw.drawWithCache
 import androidx.compose.ui.graphics.Color
-import androidx.compose.ui.graphics.drawOutline
 import androidx.compose.ui.semantics.Role
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.text.style.TextOverflow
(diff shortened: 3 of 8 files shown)


felipeemarcon (approved):
LGTM

PR #330, opened 2026-08-21
fix(topbar): make the integrated search dismissible on legacy Android

williankl:
## Summary

On Android 8.1 and below the framework re-focuses a text field right after `clearFocus` and focuses it when a window opens — on the API 27 PAX terminals the search TopBar opened with the keyboard already up and could never be dismissed.

- Dismissal now moves focus to an invisible `SearchFocusDecoy` instead of clearing it, and the decoy absorbs the focus grant when a window opens.
- The TopBar arrow hides the keyboard explicitly and is only clickable while searching — the resting magnifier stays tap-through.
- The search field's leading/clear icons get a full-height touch target, with no layout change.
- `SearchField`'s cancel button only shows while focused and no longer clears the input (KMP + SwiftUI). Hosts that want dismissal to reset the query do it in `onCancel`.
- Dismissing the keyboard with system back also unfocuses the search TopBars; the IME-inset gate drops to the compat minimum of API 23.

## API Dump

No public API changes — `bcv-check.sh --ci` verdict: **NO_CHANGES**.

## Test plan

- [x] PAX A920Pro (API 27), sample app: search opens unfocused; arrow and cancel dismiss reliably; input kept on cancel; ~44dp icon taps land
- [x] API 34 emulator: no behaviour change
- [x] ktlint, detekt, Android + desktop compile, `swift build`

### DIFF
--- a/kmp/composeApp/src/commonMain/kotlin/com/teya/lemonade/SearchFieldDisplay.kt
+++ b/kmp/composeApp/src/commonMain/kotlin/com/teya/lemonade/SearchFieldDisplay.kt
@@ -108,8 +108,8 @@ internal fun SearchFieldDisplay() {
                     )
 
                     LemonadeUi.Text(
-                        text = "Cancelling drops the focus, hides the keyboard and empties the field. " +
-                            "onCancel then runs for whatever the query was driving.",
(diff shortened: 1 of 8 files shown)


felipeemarcon (approved):
LGTM

PR #332, opened 2026-08-25
fix(button): keep the label to a single line with an end ellipsis

williankl:
# Summary
A long `Button` label used to wrap onto multiple lines, and since the button has a fixed `requiredHeight`, the extra lines were clipped by the container. Both public `Button` overloads now pass `maxLines = 1` and `overflow = TextOverflow.Ellipsis` to the label's `LemonadeUi.Text`, so an oversized label stays on one line and truncates with an ellipsis at the end.

The SwiftUI `Button` already rendered a single tail-truncated line — SwiftUI text truncates inside the fixed-height frame instead of overflowing like Compose — but only implicitly, so both label call sites now pass `maxLines: 1` to make the contract explicit and keep parity with KMP.

`IconButton` has no label and `RadioButton`'s label wraps intentionally, so neither is touched.

## API Dump
No public API changes.

https://claude.ai/code/session_01V2rKVytVxq9f7aC5TWiDzB


felipeemarcon (approved):
LGTM

PR #339, opened 2026-08-27
feat(swiftui): let a toast replace the visible one instead of queueing

andersonrsoares:
# Summary

`LemonadeToastManager.show(_:)` has always queued: a second call went into `pendingToasts` and
waited out `ToastAnimationConfig.minimumVisible` before taking over. That suits messages that each
report a distinct outcome, but not a repeated action reporting its running state — a till tapping
"add to cart" leaves a backlog that plays back after the tapping has stopped. Compose and Flutter
have no such queue, so SwiftUI was the odd one out.

Adds `LemonadeToastPolicy` and a `policy:` parameter on `show(_:)`, defaulting to `.queue`, so no
existing caller changes behaviour.

```swift
toastManager.show("Added to cart", voice: .success, policy: .replace)
```

**`.replace` clears the queue as well as taking the screen.** Superseding only the visible toast
would leave the earlier ones to be promoted when the replacement is dismissed — the backlog would
still play out, just later. It is also deliberately not routed through `dismiss()`, which promotes a
queued toast into the slot we are about to overwrite and consumes it unseen.

**The replacement adopts the visible toast's identity** (`adoptingIdentity(of:)`), so a burst reads
as one pill whose content updates rather than a run of toasts fading over each other. Going through
`displayToast` restarts the dismissal timer, so each update gets a full duration. With nothing on
screen there is no identity to adopt and the toast arrives normally.

That makes the container's render source matter. `LemonadeToastContainerView` drew a `@State` copy
refreshed only when the toast id changed — which an in-place update never does, so the pill would
keep its old text. It now draws the manager's toast whenever that carries the id already on screen,
and the captured copy otherwise. Keyed on identity rather than on `animationPhase`, because SwiftUI
computes `body` before running the `onChange` that starts an exit; phase-keyed, an incoming toast
would show for a frame before the old one faded over it.

`LemonadeToastPolicyTests` covers the replace path: queue cleared, identity adopted, timer
restarted, and `.queue` callers unaffected.

The sample app gains a queue/replace pair of burst buttons so the two can be compared side by side.

## Figma component

n/a — no visual change to the toast itself. This only changes how concurrent `show(_:)` calls
resolve against each other.

## API Dump

**Verdict:** NO_CHANGES

No public API changes — this is a SwiftUI-only change, so no `kmp/<module>/api/*.api` or
`*.klib.api` file is touched. `bcv-check.sh --ci` confirms the baseline is identical to
`origin/main`.

Worth flagging for review even though BCV does not cover it: the Swift surface does grow, and both
additions are source-compatible.

- `public enum LemonadeToastPolicy` — new type, pure addition.
- `show(_:)` gains a trailing `policy: LemonadeToastPolicy = .queue`. Defaulted, so every existing
  call site compiles unchanged and keeps today's queueing behaviour.

None.

### DIFF
--- a/swiftui/SampleApp/ToastDisplayView.swift
+++ b/swiftui/SampleApp/ToastDisplayView.swift
@@ -160,6 +160,16 @@ struct ToastDisplayView: View {
                     toastManager.show(label: "Second toast", voice: .neutral)
                     toastManager.show(label: "Third toast", voice: .error)
                 }
+
+                // Run both to compare: queued, all ten play back long after the tapping has
+                // stopped; replaced, one pill counts up and settles on "Added item 10".
+                Button("Rapid Burst (queue)") {
+                    burst(policy: .queue)
+                }
+
+                Button("Rapid Burst (replace)") {
+                    burst(policy: .replace)
+                }
             }
 
             Section("Padding") {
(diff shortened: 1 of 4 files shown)


felipeemarcon (approved):
LGTM

PR #344, opened 2026-09-01
perf(typography): resolve the font family once per theme build

### LINKED ISSUE #340: perf: lemonadeFontFamily rebuilds its Font() set on every read, multiplied by every textStyle call
## What

`lemonadeFontFamily` is a `@Composable` getter that allocates a fresh `FontFamily` with three `Font()` calls every time it is read, with no `remember`:

`kmp/ui/src/commonMain/kotlin/com/teya/lemonade/LemonadeTypographyExtensions.kt:22`

```kotlin
public val lemonadeFontFamily: FontFamily
    @Composable get() {
        return FontFamily(
            Font(LemonadeRes.font.Figtree_Regular, FontWeight.Normal),
            Font(LemonadeRes.font.Figtree_Medium, FontWeight.Medium),
            Font(LemonadeRes.font.Figtree_SemiBold, FontWeight.SemiBold),
        )
    }
```

Every `LemonadeTextStyle.textStyle` read (same file, `:34`) pulls it in, so each one costs three `Font()` composables. `Font()` from compose-resources is itself a composable that holds its own resource state, starting at an empty font file and flipping once the bytes resolve.

## Why it adds up

`lemonadeExpressiveTypography()` (`kmp/expressive/.../LemonadeExpressiveTheme.kt:55`) reads `.textStyle` **15 times** to build one Material `Typography`, so a single call is 15 `TextStyle` + 15 `FontFamily` + **45 `Font()`** composables. `.textStyle` is used in 34 further places across the codebase.

That mapper used to run once, at the app root. It now also runs per component instance: `LemonadeTimePickerTheme` scopes its own Material theme so the time picker keeps Lemonade's typeface under any host on branch `felipeemarcon/feat-time-picker`. The TimePicker sample screen has four inline pickers plus two dialogs, so the same 45-font build happens six more times on one screen.

## Suggested fix

1. Resolve the font family once and `remember` it, rather than rebuilding it per read.
2. Split the pure part out — a non-`@Composable` `LemonadeTextStyle.toTextStyle(fontFamily: FontFamily)` — so `lemonadeExpressiveTypography()` can resolve the family once and wrap the whole `Typography(...)` in a `remember`.

That collapses 45 `Font()` call sites to 3 per theme build and stops re-allocating 15 `TextStyle`s per composition. It speeds up the root theme too, not just the component-scoped ones.

## Scope note

This is pre-existing in `kmp/ui` — component-scoped Material themes multiply it rather than cause it, so it was deliberately left out of that change. Fixing it touches shared typography used by every component, which is why it is filed separately.

Found during a code-quality pass; the call counts above are read off the source, not profiled — worth measuring before and after.

felipeemarcon:
# Summary

`lemonadeFontFamily` built a fresh `FontFamily` on every read, and `LemonadeTextStyle.textStyle`
had no `remember` at all, so it rebuilt a `TextStyle` on every recomposition. `lemonadeExpressiveTypography()`
reads `.textStyle` 15 times to build one Material `Typography`, and each read is a separate call site
resolving its own copy of the family.

This remembers the resolved family and the mapped style, and has the expressive mapper resolve the
family once for all 15 styles. One theme build goes from 15 font-family resolutions to 1.

Closes #340

## Figma component

Not applicable — no component or visual change.

## Screenshot

Not applicable — rendering is unchanged; this only affects how often the font family and text styles
are rebuilt.

## API Dump

**Verdict:** ADDITIONS_ONLY

- `toTextStyle(LemonadeTextStyle, FontFamily): TextStyle` — a new function, added in all three
  baselines (`ui.api` android + desktop, `ui.klib.api`). Pure addition, so already-compiled
  consumers keep linking. It is the non-composable half of the existing `textStyle` getter, split
  out so the `expressive` module can resolve the font family once instead of once per style.
- It is marked `@InternalLemonadeApi`, since it exists to let `expressive` bypass the composable
  getter rather than as general-purpose public surface. The marker is `Level.WARNING` with `BINARY`
  retention, so it changes no symbol and the baselines are unaffected by it.

None.

## How to test

1. `cd kmp && ./gradlew :ui:apiCheck :expressive:apiCheck` — baselines match, no `apiDump` needed.
2. `./gradlew :ui:ktlintCheck :expressive:ktlintCheck detekt :ui:desktopTest` — all green.
3. Run the sample app and confirm typography renders identically: text in Figtree at the same
   sizes, weights and letter spacing, in both `LemonadeTheme` and `LemonadeExpressiveTheme`.
4. The behaviour to watch for is that nothing changed visually — the fix is purely about how often
   the font family and text styles are rebuilt.

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have added all necessary code documentation and specifications
- [ ] I have performed manual tests to ensure the system is working as expected
- [ ] I have created automated tests to replicate my manual testing automatically
- [ ] If these changes use a new version of an API, is that new version on staging and production?


kmp/expressive/src/commonMain/kotlin/com/teya/lemonade/LemonadeExpressiveTheme.kt
```diff
@@ -51,26 +52,31 @@ public fun LemonadeExpressiveTheme(
...
-        titleMedium = typography.bodyLargeSemiBold.textStyle,
-        titleSmall = typography.bodyMediumSemiBold.textStyle,
-        bodyLarge = typography.bodyLargeRegular.textStyle,
-        bodyMedium = typography.bodyMediumRegular.textStyle,
-        bodySmall = typography.bodySmallRegular.textStyle,
-        labelLarge = typography.bodyMediumMedium.textStyle,
-        labelMedium = typography.bodySmallMedium.textStyle,
-        labelSmall = typography.bodyXSmallMedium.textStyle,
-    )
+    // Resolve the family once for all fifteen styles rather than once per `.textStyle` read.
```
williankl: Noise — the code right below already says this, and `fifteen` goes stale the day a style is added. We keep this kind of rationale in the commit/PR description (where it already is 🙂).

```suggestion
```

kmp/ui/src/commonMain/kotlin/com/teya/lemonade/LemonadeTypographyExtensions.kt
```diff
@@ -14,41 +15,57 @@ import org.jetbrains.compose.resources.Font
...
+        }
     }
 
+/**
+ * Converts a [LemonadeTextStyle] to a Compose [TextStyle] using an already-resolved [fontFamily].
+ *
+ * [lemonadeFontFamily] memoises per call site, so building a set of styles through [textStyle]
+ * resolves the family once per style. Read the family once and call this instead when converting
+ * several styles together.
+ */
```
williankl, line 34: This paragraph documents `lemonadeFontFamily`'s behaviour — that can change and nobody will remember to update this doc. I'd keep the KDoc to the one-liner; the when-to-use guidance already lives in the PR description.

```suggestion
/**
 * Converts a [LemonadeTextStyle] to a Compose [TextStyle] using an already-resolved [fontFamily].
 */
```

kmp/expressive/src/commonMain/kotlin/com/teya/lemonade/LemonadeExpressiveTheme.kt
felipeemarcon: Done — dropped the line in bdd40bc1.

kmp/ui/src/commonMain/kotlin/com/teya/lemonade/LemonadeTypographyExtensions.kt
felipeemarcon: Done — KDoc trimmed to the one-liner in bdd40bc1.

williankl (approved):

PR #345, opened 2026-09-01
feat(figma): add Code Connect mappings for Compose components

felipeemarcon:
# Summary

Maps the Lemonade component library to its Compose call sites with Figma Code Connect, so Dev Mode and MCP-driven agents emit real `LemonadeUi.*` code instead of raw layer output. Today a client implementing a screen from Figma gets invented markup with hardcoded colours; this hands them the actual component and its props.

Covers **fifteen** components: Button, IconButton, Icon, TextField, SelectField, Card, ContentListItem, Tag, Checkbox, RadioButton, Toast, Chip, Link, Switch, SegmentedControl.

The templates are parserless `.figma.ts` files that emit Kotlin as strings, so **no Kotlin is added to `kmp/ui`** and the BCV baseline is untouched.

The 286 icon-library mappings are split into #346, stacked on this one. Until that lands, `icon = ...` renders as the bare instance rather than a `LemonadeIcons` value.

Already published under the `Compose` label. The unrelated `React` mappings on the same file use a separate label and are untouched.

## Mappings that needed more than a property lookup

- **SegmentedControl** — Figma numbers the selected segment from **1**; `selectedTab` is a **0-based index**, so the template converts. Verified across all 42 variants. Segment labels aren't Figma properties, so only the *count* carries over and the snippet emits `"Tab 1".."Tab n"` placeholders.
- **Toast** — the message is a plain text layer, not a property, so it's read with `findText('Label')`. Its icon is baked into the Success and Error variants; only Neutral has a real swap, so emitting it elsewhere would invent an icon.
- **Chip** — folds disabled into `Interaction State` rather than a flag, and has no slot overload, so its Figma slots map onto `leadingIcon`/`trailingIcon`.

## Deliberately unmapped

So none of it reads as an oversight:

- `Interaction State` (runtime state driven by `interactionSource`) and `Device` (no code equivalent), everywhere.
- `Card`'s heading and footer action — Compose takes `CardHeaderConfig` objects, Figma models them as nested components.
- `Link`'s `Show Indicator` — no code equivalent.
- **Country Flag** — its `◇ Flag` swap resolves to flag components (`GB-ENG-england`) with no templates, so it would emit an opaque instance where a `LemonadeCountryFlags` value belongs. The 265 flags need their own generated set, like the icons.

## Figma component

https://www.figma.com/design/91S16rhVrl5wivqV66fNjm/

## Screenshot

No visual change — this is tooling. Verified against the published mappings:

```kotlin
LemonadeUi.Button(
    label = "Button label",
    onClick = { },
    variant = LemonadeButtonVariant.Primary,
    type = LemonadeButtonType.Solid,
    size = LemonadeButtonSize.Large,
)
```

```kotlin
LemonadeUi.Toast(
    label = "Neutral message",
    voice = ToastVoice.Neutral,
    icon = LemonadeIcons.Heart,
    actionLabel = "Action",
    onAction = { },
)
```

Switching a Button instance from Large to Medium changes `size` accordingly — one template covers every variant in the set.

## How to test

```bash
cd figma && npm ci
```

Then in Figma Dev Mode, select a Button, Tag or TextField and switch the panel to the `Compose` label. Changing the instance's variant, size or disabled state should change the snippet. On a SegmentedControl, changing `Selected` should move `selectedTab` to one lower.

## API Dump

No public API changes.

### DIFF
--- a/.claude/skills/publish-figma-connect/SKILL.md
+++ b/.claude/skills/publish-figma-connect/SKILL.md
@@ -1,0 +1,165 @@
+---
+name: publish-figma-connect
+description: >
+  Publish the Lemonade Figma Code Connect mappings so Figma Dev Mode and
+  MCP-driven agents emit real LemonadeUi.* code. Use when a `figma/connect/*.figma.ts`
+  template changes, when icons are added and the templates need regenerating, when
+  a Figma component's properties are renamed and the snippets have gone stale, or
+  when the user asks to "publish Code Connect", "push the Figma mappings", or
+  "sync Figma to code".
+---
+
+# Publish Lemonade Figma Code Connect
+
+Uploads the templates in `figma/connect/` to Figma under the **`Compose`** label.
+Publishing writes to the **shared team library** — everyone in the org sees the
+result immediately. There is no staging environment.
+
+| Thing | Value |
+|---|---|
+| Config | `figma/figma.config.json` |
+| Label | `Compose` (independent namespace — publishing never touches the `React` label on the same file) |
+| Components file | `91S16rhVrl5wivqV66fNjm` |
+| Icons file | `f7zokCdnayXejxc2y7r1Qt` |
+| Token env var | `FIGMA_CODE_CONNECT_TOKEN` — **per-machine, set it up once (below)** |
+
+## First run on a new machine
+
+Nothing here is checked in — each person sets this up once.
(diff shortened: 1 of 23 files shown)


williankl (dismissed):

felipeemarcon:

PR #346, opened 2026-09-01
feat(figma): add Code Connect mappings for the icon library

felipeemarcon:
# Summary

Maps all 284 Lemonade icons, so an icon nested in a Button, Tag or IconButton resolves to a real `LemonadeIcons` value instead of an opaque instance.

**Stacked on #345** — that PR introduces the config and tooling this builds on. Review it first; the base will retarget to `main` once it merges.

288 of the changed files are generated output. The reviewable surface is three files: `scripts/generate-icon-templates.mjs`, `icons.manifest.json`, and the config/README changes.

The templates are generated rather than written by hand, mirroring how `LemonadeIcons.kt` is generated by `svg-asset-converter.main.kts`. The generator cross-checks the manifest against the enum and **fails** if the two have drifted, so a template can never point at an icon the enum does not have. Today that check is 284/284 with no orphans on either side.

## Figma component

https://www.figma.com/design/f7zokCdnayXejxc2y7r1Qt/

## Screenshot

No visual change — this is tooling. What it buys is the nested resolution. With #345 alone, a Tag's icon is an opaque instance; with this branch:

```kotlin
LemonadeUi.Tag(
    label = "Label",
    voice = TagVoice.Positive,
    icon = LemonadeIcons.Heart,
)
```

That `LemonadeIcons.Heart` is resolved out of the Icons file into a template in the Components file.

## How to test

Regenerate and confirm the tree is unchanged:

```bash
cd figma && npm ci
node scripts/generate-icon-templates.mjs
git diff --exit-code connect/icons/
```

Then validate everything parses:

```bash
```

In Figma Dev Mode, select a Tag or Button with an icon and confirm the `Compose` snippet names a real `LemonadeIcons` entry.

## API Dump

No public API changes.


felipeemarcon:

PR #348, opened 2026-09-01
feat(figma): add Code Connect mappings for SwiftUI components

felipeemarcon:
# Summary

Mirrors the Compose component mappings for the SwiftUI SDK, so an iOS engineer implementing a Figma screen gets Swift instead of Kotlin. Same ten components.

SwiftUI publishes under its own **`SwiftUI`** label from a second config, `figma.swiftui.config.json`. Labels are independent namespaces, so this doesn't touch `Compose`.

**Stacked on #346.** Merge order: #345 → #346 → this → #349.

## Do not publish the SwiftUI label from this branch alone

Figma resolves a nested icon **by node**, and when a label has no template for that node it falls back to another label's rather than rendering nothing. The SwiftUI icons are in #349, so from this branch alone a Tag renders:

```swift
LemonadeUi.Tag(
    label: "Label",
    icon: LemonadeIcons.Heart,   // Kotlin, inside a Swift call
    voice: .positive
)
```

I hit this for real while verifying, so it isn't hypothetical. `Compose` is unaffected either way. The README carries the same warning, and #349 removes it once the icons land.

## Figma component

Same components file as #345 — only the label differs.

## Screenshot

No visual change. Verified against the published `SwiftUI` label:

```swift
LemonadeUi.Button(
    label: "Button label",
    onClick: { },
    variant: .primary,
    type: .solid,
    size: .large
)
```

## Notes for reviewers

- **Leading commas.** Swift rejects a trailing comma in an argument list, so the SwiftUI templates prepend the comma to each optional argument. The Compose templates keep the trailing form Kotlin allows. `connect-swiftui/Button.figma.ts` is where an invalid snippet would show up.
- **`TextField.input` is a `Binding`**, so the snippet emits `.constant("Value")`. It keeps the designed text visible and compiles as written, but a real screen wants `@State`. Flagging in case you'd rather it emitted `$text` and dropped the designed value.
- Same properties deliberately unmapped as on Compose: `Interaction State`, `Device`, and `Card`'s heading/footer action.

## How to test

```bash
cd figma && npm ci
./node_modules/.bin/figma connect publish --config figma.swiftui.config.json \
```

## API Dump

No public API changes.

### DIFF
--- a/.claude/skills/publish-figma-connect/SKILL.md
+++ b/.claude/skills/publish-figma-connect/SKILL.md
@@ -17,8 +17,8 @@ result immediately. There is no staging environment.
 
 | Thing | Value |
 |---|---|
-| Config | `figma/figma.config.json` |
-| Label | `Compose` (independent namespace — publishing never touches the `React` label on the same file) |
+| Configs | `figma/figma.config.json` (Compose) and `figma/figma.swiftui.config.json` (SwiftUI) |
+| Labels | `Compose`, `SwiftUI` — each published separately from its own config. Neither touches the unrelated `React` label on the same file. |
 | Components file | `91S16rhVrl5wivqV66fNjm` |
 | Icons file | `f7zokCdnayXejxc2y7r1Qt` |
 | Token env var | `FIGMA_CODE_CONNECT_TOKEN` — **per-machine, set it up once (below)** |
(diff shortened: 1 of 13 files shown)


felipeemarcon:

PR #349, opened 2026-09-01
feat(figma): add Code Connect mappings for the SwiftUI icon library

felipeemarcon:
# Summary

Maps all 284 icons under the `SwiftUI` label, so a nested icon resolves to a `LemonadeIcon` case instead of falling back to Compose's Kotlin value.

**Stacked on #348.** This is the PR that makes the SwiftUI label safe to publish — see the warning in #348. Merging #348 without this one leaves `icon: LemonadeIcons.Heart` inside Swift calls.

284 of the changed files are generated. The reviewable surface is `scripts/generate-icon-templates.mjs` and the README.

## What changed in the generator

It's now parameterised per platform rather than duplicated. Two things worth checking:

- **Compose output is byte-identical.** Regenerating leaves `connect/icons/` unchanged — that's the regression check, since this refactor touches 284 already-published files.
- **SwiftUI case names are read off the enum, not derived.** `LemonadeIcon` is `String`-backed and its raw values already carry the kebab-case Figma name (`case appearanceDarkMode = "appearance-dark-mode"`), so the mapping is parsed rather than guessed by a name transform.

## Figma component

Same icons file as #346 — only the label differs.

## Screenshot

No visual change. Verified against the published `SwiftUI` label:

```swift
LemonadeUi.Tag(
    label: "Label",
    icon: LemonadeIcon.heart,
    voice: .warning
)
```

## How to test

Confirm regeneration is reproducible and Compose is untouched:

```bash
cd figma && npm ci
node scripts/generate-icon-templates.mjs
git diff --exit-code connect/icons/ connect-swiftui/icons/
```

Then validate both labels:

```bash
./node_modules/.bin/figma connect publish --config figma.config.json --dry-run --exit-on-unreadable-files
./node_modules/.bin/figma connect publish --config figma.swiftui.config.json --dry-run --exit-on-unreadable-files
```

In Figma Dev Mode, select a Tag with an icon and switch between the `Compose` and `SwiftUI` labels — you should get `LemonadeIcons.Heart` and `LemonadeIcon.heart` respectively.

## API Dump

No public API changes.


felipeemarcon:

PR #352, opened 2026-09-01
feat(web): add web as a fourth platform — design tokens

felipeemarcon:
# Summary

Adds **web as a fourth platform** to Lemonade, published as `@teya/lemonade-ds`. Scope is deliberately **design tokens only** — no React components, no Tailwind preset, no CSS reset. The component-framework decision is left until adoption shows which components are actually wanted.

Generated from the same `tokens/*.tokens.json` the KMP and SwiftUI converters read, by Kotlin `.main.kts` scripts in `scripts/`.

**Why Kotlin and not TypeScript.** The three existing loaders each carry their own copy of the DTCG parser, and `check-loader-parity.py` is what stops those copies drifting. A TypeScript loader would be a fourth copy in a language that guard cannot read — web would become the one platform able to silently disagree with the others about a token's name or value, with every platform's own tests still green. The web loader is registered with the guard, which now reports **13 shared functions identical across four loaders**.

> **Stack:** this is the base of three PRs — #352 tokens → #354 SVG assets → #353 Storybook gallery.

### What ships

| | |
|---|---|
| **Colours** | 147 semantic colours per theme. Light on `:root`, dark via `prefers-color-scheme` *and* an explicit `data-lmnd-theme` attribute that always wins and works at any depth |
| **Scales** | 122 theme-neutral custom properties — spacing, radius, size, border width, opacity, type scale, composed shadows |
| **Typography** | 30 `.lmnd-text-*` classes, extracted from the SwiftUI table rather than transcribed, with a parity check |
| **Fonts** | Figtree `.woff2` in the three weights the native platforms ship |
| **TypeScript** | `tokens` and `textStyles`, carrying **raw pixel numbers** so anything doing arithmetic gets the same values Android and iOS use |
| **Prototypes** | A self-contained `lemonade.css` for environments that cannot install from a registry, and an `llms.txt` token reference built from the `$description` guidance already in the Figma export |

Colours are deliberately **CSS-only** — a JS colour map would be a second source of truth that cannot follow the theme attribute.

### Generated vs built

Two kinds of output, and the distinction is load-bearing:

- **Generated** (`styles/*.css`, `src/*.generated.ts`, `assets/**`, `llms.txt`, `tokens.json`) is **committed**, so `token_drift.yml` can diff it. That guard is the whole justification for the Kotlin decision above.
- **Built** (`dist/**`) is gitignored. A build never modifies a committed file, so it never leaves your tree dirty.

### Nothing is published

`web_release.yml` retains `npm publish --dry-run`. Arming it needs an `NPM_TOKEN` and a decision on the scope — note Teya's public SDK uses `@teyaproduct`, so **`@teya` may already be taken on npm** and wants checking before release.

## Figma component
n/a — foundations only, no component work.

## Screenshot
n/a — no visual component in this PR. The browsable gallery is in the stacked Storybook PR.

## API Dump

No public API changes. This PR adds no Kotlin/Swift public surface; the `kmp/` and `swiftui/` trees are untouched. `scripts/web-resource-file-loading.main.kts` is a build-time script, not a published module.

---

## Two findings that need the design team

Both predate this PR — it made them visible rather than causing them.

**1. Light theme fails WCAG 2.2 AA on secondary text.** Measured with correct alpha compositing (translucent surfaces flattened over `bg-default` first), computed twice by independent implementations agreeing to ±0.02:

| Theme | Pair | Measured | Needs |
|---|---|---|---|
| Light | `content-secondary` on all four surfaces | 3.83–4.07:1 | 4.5:1 |
| Light | `content-tertiary` on all four surfaces | 2.39–2.48:1 | 4.5:1 |
| Dark | `content-tertiary` on `bg-elevated-high` | 4.34:1 | 4.5:1 |

`content-secondary` is described in the token export as *"Use for secondary text, such as body copy or supporting content."* Dark is otherwise clean. **The tokens are shared, so Android, iOS and Flutter render the same values today** — any fix belongs in Figma, not in a platform. Recorded in spec §10.

**2. `bodyMediumBold` disagrees between platforms.** SwiftUI declares `semibold`, KMP declares `Bold`. All 30 styles were audited; it is the only divergence. Web follows SwiftUI. Worth its own issue — the parity check added here guards web↔SwiftUI only, so KMP can keep drifting unwatched.

## Known gaps

- SVG assets and the Storybook gallery are stacked PRs (#354, #353); Storybook has no deploy workflow.
- The spec's Theming and "Migrating from `financial-component-library`" guides were not built.
- Contrast validation was built, run, and then withdrawn — design owns token accessibility and validates at source. The measurements it produced are preserved in spec §10.

Design doc: `.claude/specs/2026-08-21-lemonade-web-support-design.md`


felipeemarcon:
Superseded by #357 — the branch now stacks on #356 (package scaffold) rather than branching straight off `main`, and GitHub's stacked-PR support does not allow retargeting the base of a PR already in a stack. Same branch (`feat/web-tokens`), same commits, correct base.

PR #353, opened 2026-09-01
feat(web): add the Storybook token gallery

felipeemarcon:
# Summary

Adds the **Storybook token gallery** for Lemonade Web — the browsable reference that makes the tokens discoverable.

> **Stacked on #357.** Base is `feat/web-tokens`, not `main`. Review #357 first; this PR's diff is only the Storybook layer (12 files). GitHub will retarget it to `main` automatically once #352 merges.

A design system nobody can browse does not get adopted — teams reach for whatever component library they can actually see. Right now that means `financial-component-library` and `shared-internal-components`, neither aligned to Lemonade. This is the piece that changes what teams find when they look.

### What's in it

| Gallery | Contents |
|---|---|
| **Colors** | Every `--lmnd-color-*` swatch with its resolved value, in both themes |
| **Typography** | All 30 `.lmnd-text-*` styles rendered live |
| **Spacing** | The scale, measured |
| **Shadows** | The five composed elevations on an elevated surface |
| **Icons** | 286 icons, substring search, click-to-copy |
| **Installing** | Quickstart guide — install, which stylesheet to import, the theme attribute |

Each gallery **derives what it renders** from the CSSOM or the generated manifests rather than a hardcoded list, so it cannot drift from the CSS it documents.

### The theme toggle is the load-bearing bit

The toolbar switch is wired to the real `data-lmnd-theme` attribute via `withThemeByDataAttribute` — flipping it exercises the actual theming design rather than a Storybook-only background swap. Storybook's own `backgrounds` addon is disabled, since the tokens own the page background and the two would fight.

### The package stays framework-agnostic

React, Storybook and its addons are confined to `devDependencies`. `@teya/lemonade-ds` still declares **no `dependencies` and no `peerDependencies`** — v0 is framework-agnostic on purpose so a Vue app, a plain HTML prototype or an existing MUI app can consume it. React here is documentation tooling, not a consumer-facing requirement.

## Figma component
n/a — foundations gallery, no component work.

## Screenshot
Not attached — the gallery is best seen by running it:

```sh
cd web && npm install && npm run storybook   # localhost:6006
```

Worth a look at **Foundations → Colors** with the theme toggle, and **Foundations → Icons**.

## API Dump

No public API changes. No `kmp/` or `swiftui/` files are touched, and no published module gains surface.

## Not included

- **No deploy workflow.** `npm run build-storybook` produces `storybook-static/`, but nothing publishes it. Picking a host is still open.
- The spec's Radius foundation story, and its Theming and "Migrating from `financial-component-library`" guides, were not built.

### DIFF
--- a/.github/workflows/web_ci.yml
+++ b/.github/workflows/web_ci.yml
@@ -29,6 +29,7 @@ jobs:
       - run: npm run typecheck
       - run: npm test
       - run: npm run build
+      - run: npm run build-storybook
       # Proves the published tarball is installable — the gate before any publish.
       - name: Pack and install into a scratch project
         run: |
--- a/.gitignore
+++ b/.gitignore
@@ -183,4 +183,6 @@ flutter/.claude/settings.local.json
 web/node_modules/
 web/dist/
 web/*.tgz
+web/storybook-static/
 
+*storybook.log
--- a/web/.storybook/main.ts
+++ b/web/.storybook/main.ts
@@ -1,0 +1,16 @@
+import type { StorybookConfig } from '@storybook/react-vite'
+
+const config: StorybookConfig = {
+  stories: ['../stories/**/*.mdx', '../stories/**/*.stories.@(js|jsx|mjs|ts|tsx)'],
+  addons: ['@storybook/addon-essentials', '@storybook/addon-interactions', '@storybook/addon-themes'],
+  framework: {
+    name: '@storybook/react-vite',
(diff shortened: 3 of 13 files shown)


felipeemarcon:
Reopening so the PR numbers ascend with stack position — this sits fourth in the stack but had the lowest number. Same branch (`feat/web-storybook`), same commits, same base.

PR #354, opened 2026-09-01
feat(web): convert and ship the SVG assets

felipeemarcon:
# Summary

Adds the **SVG asset pipeline** for Lemonade Web: 590 assets — 286 icons, 265 flags, 39 brand logos — with a typed manifest, build-time optimization, and three subpath exports.

> **Stacked on #357.** Base is `feat/web-tokens`, not `main`. Review #357 first; this PR's diff is only the asset layer. GitHub retargets automatically as the stack merges.

## The problem this solves, and the trap in it

Every source icon is hardcoded black **twice** — as a `fill` attribute *and* an inline `style`:

```html
<path d="…" fill="black" style="fill:black;fill-opacity:1;"/>
```

An inline `style` beats every stylesheet rule, so rewriting only the attribute leaves icons rendering black no matter what CSS says. `scripts/web-svg-converter.main.kts` strips both and substitutes `currentColor`, which is what lets an icon inherit text colour and follow the theme.

**The recolouring is icon-only, and that asymmetry is the point.** All 265 flags carry inline fill styles too — but theirs are legitimate multicolour artwork. Stripping them would flatten every country flag into a grey silhouette: a defect that passes every count check and only shows up when a human looks at the page. Flags and brand logos are copied through byte-identically, verified with a full-tree diff rather than sampling.

Verified against the real asset set: icons contain only `fill="black"` (611 occurrences) and `fill="none"` (286, one per `<svg>` root) — no strokes, no gradients, no `url(#…)` references — so the rewrite provably covers every icon rather than a subset.

## Optimization

`web/build/optimize-svg.mjs` runs `svgo` into `dist/assets/`, family-aware:

- **Icons** — aggressive, with `removeViewBox` disabled (without a viewBox icons stop scaling)
- **Flags and brand logos** — conservative: `mergePaths`, `convertPathData` and `removeHiddenElems` all disabled, because aggressive path work visibly distorts multi-colour artwork

It reads the committed `assets/` and writes only into `dist/`, so **a build never mutates a committed file**.

## Consuming

```js
import ArrowRight from '@teya/lemonade-ds/icons/arrow-corner-down-left.svg'  // bundler inlines; currentColor works
import { iconNames } from '@teya/lemonade-ds'                                 // typed union
```

```html
<!-- no build step, still themeable via the .lmnd-icon mask utility from #352 -->
<span class="lmnd-icon" style="--lmnd-icon: url('…/icons/arrow-corner-down-left.svg')"></span>
```

**No sprite.** Measured: all icons together are ~123KB gzipped — that is 123KB downloaded to render one arrow. The sprite pattern solved HTTP/1.1 connection limits, which HTTP/2 removed and bundler inlining makes moot. Individual files ship only what a consumer uses.

The published tarball ships `dist/assets/**` (optimized) and not `assets/**` (converter output), so the 590 files are not duplicated in the package.

## Figma component
n/a — asset pipeline, no component work.

## Screenshot
n/a — the browsable icon gallery is in the stacked Storybook PR (#353).

## API Dump

No public API changes. No `kmp/` or `swiftui/` files are touched. `scripts/web-svg-converter.main.kts` is a build-time script, not a published module.

## Note for reviewers

The diff is ~590 files, almost all mechanical per-file rewrites. The files with actual logic are:

- `scripts/web-svg-converter.main.kts` — the two regexes and the family-aware recolouring
- `web/svgo.config.mjs` — the two optimization profiles
- `web/build/optimize-svg.mjs` — reads `assets/`, writes `dist/assets/`

Worth checking `assets/flags/` still carries inline fills, and that no `fill="black"` survives in `assets/icons/`.


felipeemarcon:
Reopening so the PR numbers ascend with stack position — this sits third in the stack but had a lower number than #356/#357, which reads confusingly. Same branch (`feat/web-assets`), same commits, same base.

PR #367, opened 2026-09-04
feat(tokens): add a themed colours layer

felipeemarcon:
Adds a generic, theme-aware colour layer — 17 hues, 170 tokens — sitting between the primitive ramps and the semantic tokens, on both KMP and SwiftUI.

**Why:** a lot of product colour carries no design-system meaning — chart series, settlement frequencies, per-role accents. Teams had two bad options: reach for `LemonadePrimitiveColors`, which is not theme-aware and so is wrong in one of the two modes, or have the design system absorb their business vocabulary, which is what the reverted #361 attempted. Applications can now pick a colour, get correct light and dark values, and supply the meaning themselves.

No semantic token is rewired here — that's a follow-up.

**The layer ships experimental.** Its shape is still settling, so KMP gates it behind `@OptIn(ExperimentalLemonadeApi::class)` at `ERROR` level — a consumer that reaches for it without opting in fails to compile. Swift has no equivalent mechanism, so it carries the same warning as documentation and is opt-in by convention there.

## Using it

Each hue carries a saturated palette and a `subtle` one. Both are a `ThemedColor`, so a component picks a variant up front and styles itself from it:

```kotlin
val style: ThemedColor = if (subtle) LemonadeTheme.themed.blue.subtle
                         else        LemonadeTheme.themed.blue

Box(Modifier.background(style.background).border(1.dp, style.border)) {
    Text(color = style.onBackground)
}

// and a colour is a value, so a chart series works
val series: List<ThemedPrimaryColor> = listOf(themed.blue, themed.amber, themed.neutral)
series.map { it.background }
```

| | `themed.blue` | `themed.blue.subtle` |
|---|---|---|
| `background` | `blue/600` → `blue/400` | `blue/α100` (both modes) |
| `backgroundHigh` | `blue/900` → `blue/200` | — |
| `onBackgroundHigh` | `blue/50` → `blue/900` | — |
| `border` | `blue/600` → `blue/400` | `blue/α100` (both modes) |
| `onBackground` | family extreme, see below | `blue/700` → `blue/400` |
| `content` | `blue/700` → `blue/400` | — |
| `contentInverse` | `blue/400` → `blue/700` | — |

```kotlin
interface ThemedColor                     { background; border; onBackground }
interface ThemedPrimaryColor : ThemedColor { content; contentInverse;
                                            backgroundHigh; onBackgroundHigh; subtle: ThemedColor }
interface BlueColors : ThemedPrimaryColor
```

Prefer a semantic token whenever one fits — both generated files say so in their header docs.

### `backgroundHigh`

A fill carrying more weight than `background`. "High" means *further from the page*, so the direction flips with the theme: `<hue>/900` in light against a white page, `<hue>/200` in dark against `neutral/900`. A fixed 900 would be invisible in dark — the dark page *is* `neutral/900`, and every hue scored 1.00–1.90 against it.

One exception: **`yellow-lime` uses `500` in dark**, not 200 — at 200 it is `#F7F7A6`, a washed-out pale that does not read as high emphasis. 500 is the step `bg-brand` uses, and still clears its label at 9.91:1.

It carries its own label. `onBackground` is tuned for `background` and fails on a 900 fill for the eight hues whose label is itself dark (amber's is `amber/950`, scoring 1.66:1). `onBackgroundHigh` takes the hue's far end from the fill instead. Worst case across all 17: fill vs page **8.67** light / **11.03** dark, label on fill **8.38** / **6.77**.

### `contentInverse`

For a surface whose brightness is inverted relative to the theme — a dark surface in light mode, a light one in dark. `content` is tuned for the default surface and fails there: on `bg-always-dark`, `content` (`blue/700`) scores **2.53:1** while `contentInverse` (`blue/400`) scores **6.56:1**.

Named for the flip rather than for "colour": it is `content` with the modes swapped, and "inverse" is already the house word for that, used across the semantic layer's Content, Background and Border groups. The semantic layer keeps `content-*-on-color` for now — renaming it is breaking, with around thirty call sites in shipped components.

## How `onBackground` is chosen

The label on a fill uses **the hue's own lightest or darkest step** where contrast allows, so a chip reads as part of its colour family.

In dark that holds for all seventeen: the background is the pale `<hue>/400`, so `<hue>/950` clears AA everywhere (worst case 5.05).

In light the background is the saturated `<hue>/600` and only ten manage it — the warm hues from the dark end, the cool ones from the light end. Seven sit too near mid-luminance for **either** end of their own ramp to reach 4.5:1; their best in-family option is 3.71–4.34. Those fall back, staying on the side their family came closest to:

- ink — orange, cyan, teal
- white — red, rose, pink, fuchsia

The subtle label is the hue's own `content`, which stays legible on a pale fill: checked across all 34 pairings, worst case 4.52.

## Light and dark

On KMP, `LemonadeTheme.themed` resolves from the existing `LemonadeSemanticColors.isDark`, so it follows whatever palette is passed to `LemonadeTheme` and there's still one brightness signal. On SwiftUI the asset catalog resolves it, as it already does for semantic colours.

One consequence: `isDark` defaults to `false`, so a custom dark palette that never overrides it gets the light themed hues. That default already existed for brand-logo assets; this gives it a second way to bite.

## Figma

The layer lives **inside the `Theme` collection** under a `Themed/` group, not in a collection of its own — `Theme` was the file's only collection carrying Light/Dark modes, and a second one would mean two mode switches per frame that can desync. Layering base onto semantic within one collection is what `Opacity`, `Radius` and `Border Width` already do.

`theme-colors.*` therefore carries both layers, and each converter filters to its own subtree: the theme converters skip `Themed/`, the themed converters keep only it.

## API Dump

**Verdict:** `ADDITIONS_ONLY` (`bcv-check.sh --ci`).

**Not a concern:** every added line is a new public type — `LemonadeThemedColors`, `ThemedColor`, `ThemedPrimaryColor`, the 17 hue interfaces, `LemonadeLightThemedColors`, `LemonadeDarkThemedColors`, `LocalThemedColors` — or the new `themed` property on the existing `LemonadeTheme` object. **Nothing was added to `LemonadeSemanticColors`**, so external palette implementors are unaffected.

**Breaking changes:** none.

## For reviewers

**1. `themed.blue.subtle` or `themed.blueSubtle`? This one needs a decision.** Nested keeps `LemonadeThemedColors` at 17 accessors however many variants appear (siblings would be 17 × variants, so 34 today), and lets a function receiving a hue reach its own variants. Both shapes cost the same 20 public types. Switching is cheap in Figma but reshapes the public interface, so it wants settling **before merge**.

**2. Adding an 18th hue is still an ABI change on paper.** The experimental marker signals instability and gives us a basis for changing the layer, but it does **not** remove the symbols from the baseline — BCV has no `nonPublicMarkers` configured, so `apiDump` still records all of it. If we want additions to be genuinely free, registering `ExperimentalLemonadeApi` as a non-public marker is the mechanism; the trade is that the dump then stops recording the layer at all, so accidental changes go unnoticed. Worth a decision either way.

**3. The SwiftUI test gives no CI signal.** `swiftui_ci.yml` runs `xcodebuild build` and no test target. `LemonadeThemedColorTests` runs only under a local `swift test`, where its asset case skips — bare SwiftPM never runs `actool`, so no asset resolves, including pre-existing semantic ones. The 136 asset lookups are verified at generation time (converter names diffed against colorset folders) and by a local `xcodebuild test`, not by CI.

**4. `neutral` is a deliberate exception.** The sixteen chromatic hues follow one step map; neutral is a mirror family — light from `neutral/alpha`, dark from `white`, at the same rung — so only `background` and `border` keep the chromatic 600→400. Its other slots take their values from the semantic tokens that already model neutral. Added so an "other" bucket in a chart stays interchangeable with the coloured hues.

**5. Two guards worth knowing about.** The converters assert that both group shapes are uniform and that `subtle`'s slots are a subset of the hue's, so a variant with its own extra slot fails the build rather than emitting types the groups can't satisfy. And `lemonadeAssetName` is deliberately shared between the two asset generators — PR #361 hit the alternative, where two converters built names differently and the lookups pointed at folders that didn't exist, failing silently at runtime.

## Follow-ups

- **PR 2 — rewiring.** Point the chromatic semantic tokens at the themed layer. Needs the loader to expose `targetVariableSetName`, which it currently discards, so converters can tell `Colors` from `Themed Colors` — a shared-DTCG-block change, so it must land in all three loaders. Carries the six approved value changes, including `content-critical`, which is at 4.47 in dark today and below AA.
- **12 leftover `Scoped/User Roles/` variables** still in the Figma `Theme` collection after the #364 revert. The next `theme-colors.*` export will bring those properties back unless they're deleted first.
- **Six `Content/Voice/On Color` colorsets with a literal space** in their folder names. Pre-existing; `lemonadeAssetName` deliberately does not slugify spaces because the generator never deletes, so fixing it means renaming assets and removing stale folders by hand.
- **`yellow` and `yellow-lime` sit very close** at every slot. Worth deciding whether both earn a place.

https://claude.ai/code/session_017CTni7CNFCnvSwisbhqVP8


felipeemarcon:
Restructured to a hue-first shape after an alignment with the team: `themed.blue.background` rather than `themed.background.bgBlue`. A hue is now a single value an app can pass around, which is what a chart series or a per-role accent needs. Values and the step map are unchanged — only names and grouping moved. Done before merge deliberately, since reshaping a public interface afterwards would be a `BREAKING` change.

That reshape surfaced a real bug in the SwiftUI themed converter, now fixed: it keyed its resource lookup on the leaf name alone, which was unique under the old grouping but repeats across hues under this one. Every token collapsed onto whichever hue sorted first, and the generated struct stopped conforming to its protocol. It now keys on the group-qualified path.

On the Copilot review:

- **Test name overclaimed** — fair, and fixed properly rather than by renaming. The test now generates all 16 × 7 names and asserts every one of the 112 assets resolves, so the name is true. Verified under `xcodebuild test`: 74 tests, 0 skipped, 0 failures.
- **Missing UIKit/AppKit import** — the suite did compile and run without them (they arrive transitively), so this wasn't a build failure. Added the explicit imports anyway, along with an `#else` branch so the file compiles on a platform with neither.

Worth flagging that the Swift test gives no CI signal: `swiftui_ci.yml` runs `xcodebuild build` and no test target, and under a bare `swift test` the asset case skips because SwiftPM never runs `actool`. The 112 lookups are verified at generation time (converter names diffed against colorset folders) and by a local `xcodebuild test` run — not by CI.

kmp/tokens/src/commonMain/kotlin/com/teya/lemonade/LemonadeThemedColors.kt
```diff
@@ -0,0 +1,205 @@
...
+
+    public interface AmberColors {
+        public val background: Color
+        public val backgroundSubtle: Color
+        public val border: Color
+        public val borderSubtle: Color
+        public val content: Color
+        public val contentOnColor: Color
+        public val onBackground: Color
+    }
```
DevSrSouza: I did not like this api design verify much, I think it should be something like:

`LemonadeThemedColors.amber`
`LemonadeThemedColors.amberSubtle`

Also second thing is that we must have a shared type across all colors. meaning:

```kotlin 
interface ThemedColor {
 public val background: Color
public val border: Color
public val content: Color
}
```

then:

```kotlin
public interface LemonadeThemedColors { 
 val amber: ThemedColor
 val amberSubtle: ThemedColor
}
```

The `contentOnColor` and `onBackground` I did not understand it, because in my mind a component that REQUIRES a ThemedColor interface as parameter for it configuration would always use the `background` so this means that the `content` color should be `onBackground` always, right? so this means have 3 types does not make sense and keep only `content` makes

kmp/tokens/src/commonMain/kotlin/com/teya/lemonade/LemonadeThemedColors.kt
```diff
@@ -0,0 +1,132 @@
...
+ * This file is automatically generated by the scripts/kmp-themed-token-converter.main.kts script.
+ * All changes will be overwritten when the script is run again.
+ *
+ * To add updated content:
+ * 1. Add your file to the tokens/ directory
+ * 2. Run: kotlin scripts/kmp-themed-token-converter.main.kts
+ *
+ * @generated by scripts/kmp-themed-token-converter.main.kts
+ */
+public interface LemonadeThemedColors {
```
williankl, line 37: I guess we also want these colors to be internal, no? or at least with the @InternalUsage annotation?

williankl (changes requested):
Looks good IMO - just internalize the coloring (or add the internal usage warning) and we should be golden

williankl (dismissed):
actually, they should be public, nevermind 😅

williankl (dismissed):

DevSrSouza (dismissed):

DevSrSouza (approved):

PR #368, opened 2026-09-04
fix(swiftui): fix unreliable top bar click with ToolbarIconButton

GustavoASantos:
# Summary
Clicking the edge of an ios Button with a lemonade icon did not respond, including in lemonade's topbar due to a weird interaction between swift ui and uikit, see claude's explanation below.

An icon-only button in a `ToolbarItem` responds only across the middle of its circle. SwiftUI hosts a bar item whose content it cannot recognise in a `UIKitBarItemHost`, capped at 36x36 with a 24x36 interactive view inside it, so roughly 6pt of each edge is dead while the control still looks like a comfortable target. Measured on an iPhone by dumping the nav bar's view tree:

  bare icon label     UIKitBarItemHost    {{337, 63}, {36, 36}}
    live area within  _UIScrollPocket     {{343, 63}, {24, 36}}
  wrapped in a Label  _UIButtonBarButton  {{337, 63}, {36, 36}}

Nothing applied to the label moves those numbers — a larger `.frame`, a `.contentShape`, and a button style were each measured and left the interactive view at 24pt, because UIKit clips hit-testing to the bar item's bounds. What does change it is the label's shape: wrapped in a `Label`, SwiftUI hands UIKit a real `_UIButtonBarButton`, which hit-tests its whole frame. `.iconOnly` means this renders exactly as the bare image everywhere else.

`lemonadeTopBar`'s own close buttons also drop `clickableFrame(minSize: .size800)`. At 32pt it is under the 36pt cap so it never grew anything, and applying it to the label is itself part of what forces the hosted path.

## API Dump

**Verdict:** 

### DIFF
--- a/swiftui/SampleApp/HomeView.swift
+++ b/swiftui/SampleApp/HomeView.swift
@@ -166,12 +166,11 @@ struct HomeView: View {
         .disableAutocorrection(true)
         .toolbar {
             ToolbarItem(placement: .topBarTrailing) {
-                Button(action: { showSettings = true }) {
-                    LemonadeUi.Icon(
-                        icon: .gear,
-                        contentDescription: "Settings"
-                    )
-                }
+                LemonadeUi.ToolbarIconButton(
+                    icon: .gear,
+                    contentDescription: "Settings",
+                    action: { showSettings = true }
+                )
             }
         }
         .sheet(isPresented: $showSettings) {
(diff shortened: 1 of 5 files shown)


ujizin (approved):

felipeemarcon (approved):
LGTM

PR #369, opened 2026-09-07
fix(button): align button spacing and icon-button glyphs with Figma

williankl:
## Summary

Audited every button size in code against the Figma component sets — **Button** (node `8302:10112`) and **Icon Button** (node `17383:2038`) — reading each size variant's geometry and variable bindings. Heights, radii, paddings, and text styles all match token-for-token on KMP and SwiftUI; three deviations were found and are fixed here:

- **XSmall label inner padding** — Figma binds `spacing-100` (4dp) for XSmall, `spacing-200` (8dp) for the other sizes; code used a flat 8dp everywhere. `LemonadeButtonContentData` now carries a per-size `labelPadding` on both platforms.
- **IconButton Medium glyph** — Figma pairs the 48dp box with a 20dp glyph; both platforms rendered 24dp. Note: Medium is the IconButton default size, so the glyph visibly shrinks wherever it's used.
- **IconButton Small glyph (SwiftUI only)** — was 16dp; Figma and KMP use 20dp (KMP was aligned in #297, which never got its SwiftUI twin).

Documented but intentionally not changed: the 64dp min-width on labeled buttons (tap-target floor; Figma is hug-width), code's XSmall IconButton (no Figma counterpart), and the Flutter button drift (xSmall height 40 vs 32, medium radius 12 vs 14, xSmall radius 8 vs 10) — left for a separate pass.

## Before / after (Android emulator)

Captured from `:composeApp` debug builds of the base commit vs this branch on a Pixel 7 AVD (API 36), same navigation sequence. The pixel diff between builds is confined to the expected regions: the Button sizes row (XSmall tightens by 4dp per side) and the Medium icon-button glyph column (48→40px on-screen, exactly the 24→20dp ratio).

**Button — XSmall label padding** (XSmall narrows, other sizes shift left):

Button before/after

**IconButton — Medium glyph 24→20** (box sizes unchanged):

IconButton before/after

The SwiftUI-only Small glyph fix is not visible in these Android captures.

*Screenshots are hosted on the `pr-369-assets` branch — delete it after merge.*

## API Dump

No public API changes. `bcv-check.sh --ci` verdict: `NO_CHANGES` — all touched declarations (`LemonadeButtonContentData`, `IconButtonSizeData`, and their SwiftUI twins) are private.

https://claude.ai/code/session_019tgToF9ENn3TNJ91Vy3WBQ

### DIFF
--- a/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/Button.kt
+++ b/kmp/ui/src/commonMain/kotlin/com/teya/lemonade/Button.kt
@@ -106,7 +106,7 @@ public fun LemonadeUi.Button(
                     textAlign = TextAlign.Center,
                     maxLines = 1,
                     overflow = TextOverflow.Ellipsis,
-                    modifier = Modifier.padding(horizontal = LocalSpaces.current.spacing200),
+                    modifier = Modifier.padding(horizontal = size.contentData.labelPadding),
                 )
 
                 if (trailingIcon != null) {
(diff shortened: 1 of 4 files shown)


felipeemarcon (approved):
LGTM

PR #373, opened 2026-09-07
fix(swipe-action-row): hand back a drag claim a context menu took

### LINKED ISSUE #370: SwipeActionRow keeps its drag claim when a context menu takes the touch
### What happens

A `LemonadeUi.SwipeActionRow` whose content also carries a `.contextMenu` is left visibly offset after the menu is dismissed, then springs back into place.

Measured frame by frame from a screen recording on an iPhone 17 Pro Max (iOS 26): the row's content — avatar, label and chevron together — sits **22px (7.3pt) to the left** for **~0.55s** after dismissal, then settles.

It is a uniform translation of the row's content, not a scale or a layout change. The label and support text measure byte-identical widths in both states (269px and 491px), while the card, the divider and the row's own frame never move.

### Why

In `swiftui/Sources/Lemonade/Components/LemonadeSwipeActionRow.swift`:

```swift
DragGesture(minimumDistance: claimDistance)   // claimDistance = 24
    .updating($isDragging) { _, state, _ in state = true }
```

```swift
.onChange(of: isDragging) { dragging in
    // A cancelled gesture never delivers `onEnded`, so the snap back has to happen here.
    guard !dragging, dragOrigin != nil else { return }
    dragOrigin = nil
    committed = false
    withAnimation(settle()) {
        travel = open ? restingTravel : 0
        holding = false
    }
}
```

Holding a row to raise a context menu drifts the finger past the 24pt claim distance, so the drag claims the touch and nudges `travel` open. The context menu then takes the touch, so `onEnded` never arrives.

`isDragging` is a `@GestureState`, and it only clears when the menu finally releases the touch — **on dismissal**. The cancellation path in `onChange(of: isDragging)` is correct in itself, but it fires far too late: the reader sees the row still offset when the menu goes away, and then watches `settle()` animate it home.

### Expected

A press that becomes a long press should release the drag's claim, so the row is back at rest by the time the menu appears.

### Reproduce

A `SwipeActionRow` with one action, its content carrying a `.contextMenu`. Press and hold with normal finger drift, then dismiss the menu by tapping away.

### Notes

The row owns the gesture with:

```swift
.highPriorityGesture(drag, including: enabled && !actions.isEmpty ? .all : .subviews)
```

so a call site cannot opt out per-gesture — neither `claimDistance` nor a per-touch disable is reachable from outside the component. Working around it means dropping one of the two affordances.

Seen in [teya-business-app#4483](https://github.com/saltpay/teya-business-app/pull/4483), where saved-account rows offer both swipe-to-remove and hold-to-remove.

felipeemarcon:
# Summary

Closes #370.

A `SwipeActionRow` whose content also carries a `.contextMenu` was left visibly offset after the menu was dismissed, then sprang back. Holding the row drifts the finger past the 24pt claim distance, so the drag claims the touch and nudges the row open; the menu then takes the touch, so the drag never ends. `isDragging` is a `@GestureState` and only clears on dismissal — the cancel path was correct but fired far too late.

The row now times the claim itself rather than waiting on a gesture that may never end. 0.3s after a drag claims the row, if the row has been carried less than 24pt since, the claim goes back and the row settles. A finger still on the row claims again on its next move, from where the row is being drawn, so a slow swipe carries on unaffected.

The rule is not context-menu-specific: a claim that took the row nowhere the reader can see was not a swipe. Anything else that steals a touch — a drag session, a long-press sheet — gets the same treatment without being named.

## Figma component
- [Design](https://www.figma.com/design/91S16rhVrl5wivqV66fNjm/%F0%9F%8D%8B-Lemonade-DS---Components?node-id=21709-6930&t=pJs6USLSPXUzgns6-11)

## Evidences
https://github.com/user-attachments/assets/d7958e31-ea5e-4b76-af11-fe7788ede5a2

## API Dump

No public API changes.

## How to test

1. Sample app → **SwipeActionRow** → the **Swipe or hold** card near the bottom. The row carries a `.contextMenu`.
2. Press and hold with normal finger drift → the menu appears; dismiss it by tapping away → the row is already at rest, with no spring-back.
3. Swipe the row open → still opens and settles normally.
4. Tap the row → its action still fires.
5. Swipe out, then pause with your finger still down → the row stays where you left it and does not snap shut.
6. Scroll the list → unaffected.

## Checklist:

- [x] I have performed a self-review of my own code
- [x] I have added all necessary code documentation and specifications
- [x] I have performed manual tests to ensure the system is working as expected
- [x] If these changes use a new version of an API, is that new version on staging and production?

### DIFF
--- a/swiftui/SampleApp/SwipeActionRowDisplayView.swift
+++ b/swiftui/SampleApp/SwipeActionRowDisplayView.swift
@@ -314,49 +314,7 @@ struct SwipeActionRowDisplayView: View {
                         }
                     }
 
-                    // Both affordances on one row: the swipe and the hold. A press drifts far
-                    // enough to claim the drag on its way to the menu, and the row has to give
-                    // that claim back rather than sit offset behind the menu until it goes.
-                    LemonadeUi.Card(
-                        header: CardHeaderConfig(
-                            title: "Swipe or hold",
-                            subtitle: "The row carries a context menu. Holding it leaves it where it is."
-                        )
-                    ) {
-                        LemonadeUi.SwipeActionRow(
-                            actions: [
-                                LemonadeSwipeAction(icon: .trash, contentDescription: "Delete", onClick: { })
-                            ],
-                            allowsFullSwipe: false
-                        ) {
-                            LemonadeUi.ActionListItem(
-                                label: "Swipe or hold",
-                                supportText: "Both reach Delete",
-                                showDivider: false,
-                                onItemClicked: { }
-                            )
-                        }
-                        // Around the row rather than inside it. The lift a context menu does is a
-                        // snapshot of the view it is attached to, and a list item's own fill is
-                        // clear — so something has to give the snapshot a surface. Putting that
-                        // surface in the content would bury the row's own press tint, which is
-                        // translucent and drawn behind whatever the content is: the row would
-                        // stop filling in under a swipe, alone among the rows on this screen.
-                        .background(
-                            RoundedRectangle(cornerRadius: LemonadeTheme.radius.radius500)
-                                .fill(LemonadeTheme.colors.background.bgDefault)
-                        )
-                        // The fill alone is not enough: the preview is otherwise clipped to iOS's
-                        // own corner radius, so the lift starts on one shape and settles on
-                        // another, and the row flickers as it goes.
-                        .contentShape(
-                            .contextMenuPreview,
-                            RoundedRectangle(cornerRadius: LemonadeTheme.radius.radius500)
-                        )
-                        .contextMenu {
-                            Button("Delete", role: .destructive) { }
-                        }
-                    }
+                    swipeOrHoldCard
 
                     LemonadeUi.Card(
                         header: CardHeaderConfig(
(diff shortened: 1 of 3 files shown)


swiftui/Sources/Lemonade/Components/LemonadeSwipeActionRow.swift
```diff
@@ -632,15 +662,42 @@ struct LemonadeSwipeActionRowView<Content: View>: View {
...
+        // and is the wrong one: a long press that recognises first takes the touch off the
+        // context menu the row is trying to stay out of the way of.
+        .task(id: dragOrigin) {
+            guard let origin = dragOrigin else { return }
+            do {
+                try await Task.sleep(nanoseconds: holdDelayNanos)
+            } catch {
+                return
             }
+            guard swipeHoldReleasesClaim(travelSinceClaim: travel - origin) else { return }
```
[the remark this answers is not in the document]
felipeemarcon: Fixed in 298391a2 — the point holds, and there was a sharper version of it: `origin` was captured before the sleep and never re-checked, so a stale timer could hand back a *different* drag's claim using the old origin's arithmetic. The post-sleep continuation now guards on both `!Task.isCancelled` and `dragOrigin == origin`.

`releaseClaim()` did already guard `dragOrigin != nil`, so a claim that had simply ended was safe; the replaced-claim case was not. The `settleOrigin = nil` you cite is gone as well — it was removed in 8c05074a for an unrelated reason (it discarded a legitimate settle when a reader paused mid-close).

swiftui/SampleApp/SwipeActionRowDisplayView.swift
```diff
@@ -245,6 +245,43 @@ struct SwipeActionRowDisplayView: View {
...
+                            // attached to, and a list item's own fill is clear — the Card behind
+                            // it is the surface. Without one of its own the row lifts as floating
+                            // text, so this gives it the fill and the shape the row already wears
+                            // under a finger.
+                            .background(
+                                RoundedRectangle(cornerRadius: LemonadeTheme.radius.radius500)
+                                    .fill(LemonadeTheme.colors.background.bgDefault)
+                                    .padding(LemonadeTheme.spaces.spacing100)
+                            )
+                            .contextMenu {
```
matheuslutero: The UI is flickering because the context menu starts with a different corner radius.
Adding a `.background(RoundedRectangle)` doesn't solve it.

Image

Please try adding this before `.contextMenu`:
```
.contentShape(.contextMenuPreview, RoundedRectangle(cornerRadius: LemonadeTheme.radius.radius500))
```

felipeemarcon: Good catch, applied in e1214eae. `contextMenuPreview` is iOS 15 / macOS 12, so it clears the package floor with no availability guard.

One adjustment on top of your line: I dropped the `.padding(spacing100)` from the fill. The `contentShape` you suggested covers the full bounds, so an inset fill would have left an 8pt transparent gutter inside the preview — the fill and the preview shape have to be the same rectangle, not just the same radius. In place it makes no difference either way, since `bgDefault` on a `bgDefault` card is invisible.

https://github.com/user-attachments/assets/11ce80f4-7178-44f6-a366-6a2214a104a1

Worth flagging the underlying issue separately: the call site should not need to know the row's corner radius to get a correct lift. Between the fill, the shape and the gutter, that is three internal details of `SwipeActionRow` hand-copied into the sample, and they will drift the next time the row's surface is retuned. If a context menu on row content is a supported composition now, `SwipeActionRow` should provide the preview surface itself. Let's do that as a follow-up rather than widen this PR.

felipeemarcon: Follow-up on this one, in 498b04ca — the fix above was right about the radius but the surface it sat on caused a second problem.

`bgSubtleInteractive` is translucent (alpha 0.1 light, 0.2 dark), which is why the row draws it *behind* the content: it tints whatever surface is underneath. The opaque fill the lift needs was sitting between the tint and the text, so this row stopped filling in under a swipe — the only row on the screen that did not.

The call site cannot get beneath the row's tint, so the `.background`, the `.contentShape` and the `.contextMenu` now go around the row rather than inside its content closure. The content is clear again, the tint behaves like every other row, and the snapshot the lift takes is the whole row. Verified on device: swipe fills in, hold lifts on the right radius with no flicker.

This is the third internal detail of `SwipeActionRow` the call site had to know to get a correct lift — the fill, the radius, and now the layering. Still happy to make the row own the preview surface as a follow-up.

felipeemarcon: Raised as #387, with the three details written up and the constraint that any fix has to leave the press tint alone.

felipeemarcon: @matheuslutero I've created the issue #387 for that

matheuslutero (approved):

PR #379, opened 2026-09-08
fix(topbar): hold the collapse position as a fraction, not a pixel offset

RomanTruba-Teya:
# Summary

`TopBarState` seeded its scroll offset from `maxScrollOffset`, which the layout does not report until it has measured the collapsable slot. Two bugs followed:

- A bar built with `startCollapsed = true` read as fully expanded for the frame before the measurement landed, so it painted its large title once and then snapped shut.
- The `Animatable` lived inside a `derivedStateOf` that read `maxScrollOffset`, so every later remeasure rebuilt it from scratch and threw away wherever the user had scrolled the bar to.

The position is now held as a `0f..1f` fraction, which is meaningful before any measurement and survives one: `collapseProgress` reads it directly, `scrollOffset` multiplies it by the measured height, and the nested-scroll handlers convert their pixel deltas through it. `collapse()` / `expand()` animate the fraction to `1f` / `0f` instead of chasing a pixel target that may not exist yet.

Adds `TopBarStateTest` covering both bugs: the pre-measurement read for either `startCollapsed` value, that the first measurement doesn't move the bar, and that a remeasure to a different height keeps the bar at the same fraction.

## Figma component

n/a — behavioural fix, no visual change to the component itself.

## Screenshot

| Before | After |
|--------|--------|
| <video src="https://github.com/user-attachments/assets/4489620f-e392-42d9-b71b-ce68ad6cd911" /> | <video src="https://github.com/user-attachments/assets/9e4228cb-4773-4312-90b6-ef7cbea8060a" /> | 

## API Dump

No public API changes.

**Verdict:** NO_CHANGES

`.claude/skills/binary-compatibility/scripts/bcv-check.sh --ci` → `NO_CHANGES — public API is identical.` No `api/*.api` or `*.klib.api` file is touched by this branch. `collapseProgress` keeps its type and semantics; everything that changed shape (`scrollOffsetAnimatable` → `collapseFraction`, `asCollapseFraction`) is `private`.

https://claude.ai/code/session_01DM9JatcxYNLemL6SXmNJxh


williankl:
Rebased onto main (`92ef689c`, force-push) to clear the conflict with #406, which had reworked the same region of `TopBarState` in the meantime.

The resolution keeps both fixes: the bar's position is now the `collapseFraction` from this PR, while #406's invariant is preserved — `currentCollapseProgress` stays a plain internal getter (now reading the fraction directly, still coerced to `0f..1f`) and the `graphicsLayer` block is untouched. Since the fraction is meaningful before measurement and survives a remeasure, #406's Animatable recreation and the `Snapshot.withoutReadObservation` guard on `maxScrollOffset` became unnecessary and were dropped for a plain `mutableFloatStateOf`. `TopBarStateTest` is unchanged and passing; BCV still reports `NO_CHANGES`.

williankl (approved):

felipeemarcon (approved):
LGTM

PR #380, opened 2026-09-08
feat(figma): connect Toast, Chip, Link, Switch and SegmentedControl

felipeemarcon:
# Summary

Extends the Compose mappings with the five most-used components that were still unconnected. All five exist on both platforms; SwiftUI equivalents would be a follow-up.

Picked by cross-referencing every Figma component against both codebases: 21 components exist on both platforms and aren't mapped. These are the top of that list by library usage, minus the ones with structural blockers.

**Stacked on #349.** Merge order: #345 → #346 → #348 → #349 → this.

## Three needed more than a property mapping

**SegmentedControl** — Figma numbers the selected segment from **1**; `selectedTab` is a **0-based index**. The template converts, so `Selected=5` emits `selectedTab = 4`. Verified across all 42 variants. The segment labels aren't Figma properties, so only the *count* carries over and the snippet emits `"Tab 1".."Tab n"` placeholders.

**Toast** — the message is a plain text layer, not a property, so it's read with `findText('Label')`. Its icon is baked into the Success and Error variants; only Neutral exposes a swappable one, so emitting the swap on the others would invent an icon that isn't there.

**Chip** — folds the disabled state into `Interaction State` rather than exposing a flag, and has no slot-based overload, so its Figma slots map onto `leadingIcon`/`trailingIcon`.

## Why Country Flag isn't here

It was in the batch of six I proposed, and I dropped it. Its `◇ Flag` swap pulls in flag components (`GB-ENG-england`) that have no templates, so `flag = ...` would emit an opaque instance where a `LemonadeCountryFlags` value belongs — the same failure that bit the SwiftUI icons in #348. The **265** flags need their own generated set, like the icons. Happy to do that as a separate PR.

## Figma component

Same components file. Nodes: Toast `7115:77429`, Chip `18177:1701`, Link `4365:4446`, Switch `5144:128`, SegmentedControl `11526:16665`.

## Screenshot

No visual change. Verified against the published `Compose` label:

```kotlin
LemonadeUi.Toast(
    label = "Neutral message",
    voice = ToastVoice.Neutral,
    icon = LemonadeIcons.Heart,
    actionLabel = "Action",
    onAction = { },
)
```

```kotlin
LemonadeUi.SegmentedControl(
    properties = listOf(
        TabButtonProperties.label(label = "Tab 1"),
        TabButtonProperties.label(label = "Tab 2"),
        TabButtonProperties.label(label = "Tab 3"),
    ),
    selectedTab = 2,
    onTabSelected = { },
    size = LemonadeSegmentedControlSize.Large,
)
```

## Notes for reviewers

- `Link`'s `Show Indicator` has no code equivalent and is unmapped.
- `Interaction State` and `Device` remain unmapped throughout, as on the earlier batch.
- The cross-reference also surfaced design/code drift worth a separate conversation: `Badge`, `Spinner`, `PinCode` and `SelectListItem` exist in code with no Figma component of that name; `Divider` and `Slider` exist in Figma with no code on either platform.

## How to test

```bash
cd figma && npm ci
```

In Dev Mode, select a SegmentedControl and change its `Selected` variant — `selectedTab` should track it, one lower.

## API Dump

No public API changes.


felipeemarcon:
Folded into #345, which is where the Compose component mappings live. The five templates (Toast, Chip, Link, Switch, SegmentedControl) and the README notes are now part of that PR — nothing is lost, and the stack is one branch shorter.

PR #381, opened 2026-09-08
feat(figma): set up Figma Code Connect

felipeemarcon:
# Summary

Adds the toolchain and configuration that the next two PRs hang templates off, so the mapping work arrives as reviewable batches instead of one large drop.

Two configs, one per platform, each with its own label — the structure [Figma documents for multi-framework repos](https://developers.figma.com/docs/code-connect/api/config-file/).

Templates are parserless and emit Kotlin or Swift as strings, so **nothing is added to `kmp/ui` or `swiftui/Sources`** and neither API baseline moves.

**First of three.** #382 adds the assets, #383 the components.

## Two behaviours worth knowing

- **`--dry-run` still needs a token.** It writes nothing, but exits 1 without `FIGMA_ACCESS_TOKEN`, so CI needs it as a secret.
- **A config whose glob matches zero templates is an error**, not a no-op. Validation therefore starts working with #382, which is the first PR to add templates.

Both are documented in the README and the publish skill.

## How to test

```bash
cd figma && npm ci
```

Then confirm the foot-gun is closed — this should error rather than publish:

```bash
./node_modules/.bin/figma connect publish --dry-run
```

## API Dump

No public API changes.


figma/README.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+Publishing writes to the shared team library. **Publish both labels**, and never
+a platform's components without its assets — Figma resolves a nested icon by
+node, and a label with no template for that node falls back to another label's,
+so a missing SwiftUI icon renders the *Kotlin* snippet inside a Swift call.
+
+The components file also carries an unrelated `React` label pointing at a
+personal exploration repo. Labels are independent namespaces; publishing these
+two does not touch it.
```
williankl: Small thing, but this repo is public — "personal exploration repo" is the sort of detail that's useful in a Slack thread and a bit odd in a README strangers read. The load-bearing part is just that there's a third label we don't own.

```suggestion
The components file also carries a `React` label published from outside this
repo. Labels are independent namespaces; publishing these two does not touch
it.
```

figma/README.md
```diff
@@ -0,0 +1,68 @@
...
+Binary Compatibility Validator baseline.
+
+File keys live only in `documentUrlSubstitutions`; templates reference
+`<LEMONADE_COMPONENTS>`, `<LEMONADE_ICONS>` and `<LEMONADE_FLAGS>`.
+
+## Validate
+
+```bash
+cd figma && npm ci
+FIGMA_ACCESS_TOKEN=figd_... npm run validate
```
williankl: Inline assignment puts the literal token on the command line, so it lands in `~/.zsh_history`. The skill already does this the right way (export once, reference it) — worth matching here so nobody copy-pastes the leaky form.

```suggestion
FIGMA_ACCESS_TOKEN="$FIGMA_CODE_CONNECT_TOKEN" npm run validate
```

Same for the two `publish:` lines further down.

.claude/skills/publish-figma-connect/SKILL.md
```diff
@@ -0,0 +1,165 @@
...
+| Token env var | `FIGMA_CODE_CONNECT_TOKEN`, set up once per machine (below) |
+
+## First run on a new machine
+
+Nothing here is checked in. Each person sets it up once.
+
+1. **Node 18+** (`node -v`). The CLI declares `engines.node >= 18`.
+2. **Install**: `cd figma && npm ci`. `@figma/code-connect` comes from public
+   npm. If your npm points at Teya's JFrog registry, it must proxy npmjs for this
+   to resolve; `npm config get registry` shows where you point.
```
williankl: Same public-repo thought as the README one. This is the only mention of our internal registry anywhere in the repo, and naming it doesn't really help the reader — "your registry has to proxy npmjs" covers the same failure.

```suggestion
2. **Install**: `cd figma && npm ci`. `@figma/code-connect` comes from public
   npm, so your registry has to proxy npmjs for this to resolve;
   `npm config get registry` shows where you point.
```

Line 37's "Teya is on `org`" is the same category — the Figma seat requirement stands fine on its own.

williankl (dismissed):
Approving. The two-config split is right, and the "deliberately not named `figma.config.json`" note is a good catch — a bare `publish` silently doing one platform and reporting success is exactly the kind of thing that bites six months later.

Only comments are public-repo hygiene: a few lines written for an internal audience that read a bit oddly on a repo strangers browse. None of it blocks.

One for whoever merges the stack rather than for this PR: #382 can't land on its own (comment over there), so these want to go in order.

figma/README.md
felipeemarcon: Taken, wording applied as suggested — b71d795. Agreed the label owner is not the load-bearing part.

figma/README.md
felipeemarcon: Fixed in b71d795: the README now uses `FIGMA_ACCESS_TOKEN="$FIGMA_CODE_CONNECT_TOKEN"` for validate and both publish lines, matching the skill.

.claude/skills/publish-figma-connect/SKILL.md
felipeemarcon: Applied in b71d795, both the registry sentence and the plan tier. The failure row in the table already says "a registry that does not proxy public npm", so nothing is lost.

williankl (approved):
Re-reviewed after the push — all three are applied and the new wording reads well. The failure table still carries the registry case without naming the box, so nothing's lost by dropping it from the setup step.

Approving. This one's ready.

PR #382, opened 2026-09-08
feat(figma): connect the icon and country flag libraries

felipeemarcon:
# Summary

Maps all **286 icons** and **264 country flags** on both platforms, so an asset nested in a Button, Tag or Country Flag resolves to a real enum value instead of an opaque instance.

**Stacked on the setup PR.** Assets come before components deliberately — see below.

1,100 of the changed files are generated. The reviewable surface is `scripts/generate-asset-templates.mjs` and the two manifests.

## Why assets come before components

Figma resolves a nested asset **by node**, and when a label has no template for that node it falls back to another label's rather than rendering nothing. Publishing a platform's components before its assets doesn't degrade gracefully — it produces confidently wrong output. I hit this for real earlier in this work: SwiftUI components without SwiftUI icons rendered

```swift
icon: LemonadeIcons.Heart   // Kotlin, inside a Swift call
```

Assets have no equivalent downside: an asset template depends on nothing and is correct standalone.

## The generator

One script covers both asset types across both platforms:

```bash
node scripts/generate-asset-templates.mjs                # everything
node scripts/generate-asset-templates.mjs flags swiftui  # one of each
```

It cross-checks each manifest against that platform's enum **in both directions**. Only checking manifest → enum catches a deleted asset but stays silent on an added one — which is the direction that actually happens, and did: `moon` and `sun` shipped in code with no mapping and nothing failed.

Swift case names are read off the enum rather than derived: `LemonadeIcon` and `LemonadeCountryFlag` are `String`-backed and their raw values already carry the Figma name.

A code entry with no Figma component goes in the manifest's `knownUnmapped` list, so the gap is visible rather than silently tolerated. Today that is one flag, `CD-congo-democratic-republic`.

## Screenshot

No visual change. Verified against both published labels:

```kotlin
LemonadeCountryFlags.ACAscensionIsland    // Compose
```
```swift
LemonadeCountryFlag.aCAscensionIsland     // SwiftUI
```

## How to test

```bash
cd figma && npm ci
node scripts/generate-asset-templates.mjs
git diff --exit-code connect/ connect-swiftui/

FIGMA_ACCESS_TOKEN=figd_... npm run validate
```

## API Dump

No public API changes.


figma/icons.manifest.json
```diff
@@ -0,0 +1,291 @@
...
+    "basket": "38:4763",
+    "basket-plus": "38:4764",
+    "basket-remove": "38:4765",
+    "battery-half": "38:4766",
+    "bell": "38:4767",
+    "blocks-fitting": "4008:36",
+    "bluetooth": "4420:43",
+    "board": "4008:38",
+    "brand-android": "3233:58",
+    "brand-apple": "3233:59",
```
williankl, line 47: `BrandTeyaPos` landed on main back in #393, but there's no `brand-teya-pos` entry here — so `node scripts/generate-asset-templates.mjs` exits 1 on this branch. The bidirectional check catching a real gap, which is reassuring, but it does mean this PR doesn't reproduce from its own generator.

#383 adds the entry and the template, so the stack as a whole is fine. Flagging mainly so it doesn't get merged on its own.

figma/icons.manifest.json
```diff
@@ -0,0 +1,291 @@
+{
+  "fileKey": "f7zokCdnayXejxc2y7r1Qt",
```
williankl: `fileKey` doesn't look like it's read by anything — the generator resolves URLs through `asset.urlToken` and `documentUrlSubstitutions`, and I couldn't find another reader. If that's right it'll drift away from the config with nothing noticing. Drop it?

It also quietly makes the README's "File keys live only in `documentUrlSubstitutions`" untrue — counting all three manifests and the skill's table they're in six files. Not a security thing (the keys are already on main in `export-icons` and `figma-themed-component`), but if anyone ever acts on that sentence they'd scrub two files and miss four.

figma/scripts/generate-asset-templates.mjs
```diff
@@ -0,0 +1,177 @@
...
+        console.error(`[${assetName}/${platformName}] ${missingFromEnum.length} Figma component(s) have no enum entry:`)
+        for (const n of missingFromEnum) console.error(`  ${n}`)
+        console.error(`  → run the svg-asset-converter to add them to ${enumPath}`)
+      }
+      if (missingFromManifest.length) {
+        console.error(`[${assetName}/${platformName}] ${missingFromManifest.length} enum entr(ies) have no mapping:`)
+        for (const m of missingFromManifest) console.error(`  ${m}`)
+        console.error(`  → refresh ${asset.manifest} from Figma, or add the Figma name to its knownUnmapped list`)
+      }
+      continue
```
williankl: Since the validation and the `rmSync`/write both sit inside the platform loop and the failure path is `continue`, a run where the Kotlin and Swift enums are out of step regenerates one platform and skips the other — which is the mixed-language state the comment at the top of this file warns about.

Could the check run for every (asset, platform) pair before the first `rmSync`, and abort rather than continue? WDYT?

figma/scripts/generate-asset-templates.mjs
```diff
@@ -0,0 +1,177 @@
...
+for (const assetName of assets) {
+  const asset = ASSETS[assetName]
+  const manifest = JSON.parse(readFileSync(join(root, asset.manifest), 'utf8'))
+  const items = manifest[asset.entries]
+  // Entries the code enum has but Figma does not, stored as Figma names so each
+  // platform maps them into its own key space below.
+  const knownUnmapped = manifest.knownUnmapped ?? []
+  // Enum entries served by another entry's component, keyed Figma name -> the
+  // Figma name that covers it.
+  const aliases = manifest.aliases ?? {}
```
williankl: Only `Object.keys(aliases)` gets used, so the values are read by nothing and this ends up behaving identically to `knownUnmapped`. No manifest uses it today either.

The README describes it as though the covering component's node gets a template pointing at the served entry, which isn't what happens. Either wire the value up or drop the concept and the docs around it — right now it reads like a feature that exists.

figma/scripts/generate-asset-templates.mjs
```diff
@@ -0,0 +1,177 @@
...
+
+export default {
+  example: figma.${platform.tag}\`${type}.${memberFor.get(name)}\`,${
+    imports.length ? `\n  imports: [${imports.map((i) => `'${i}'`).join(', ')}],` : ''
+  }
+  id: '${assetName.replace(/s$/, '')}-${name}',
+  metadata: { nestable: true },
+}
+`
+      writeFileSync(join(outDir, `${pascal(name)}.figma.ts`), body)
```
williankl: Two small ones here.

`pascal()` collapses `-`, `_` and spaces, so two manifest keys could resolve to the same filename and the second would silently win. The count on line 172 comes from `readdirSync` rather than `Object.keys(items).length`, so it'd just print a smaller number and nobody would notice. `figma-icons.py:88-92` already dies on exactly this case — might be worth stealing.

Separately, per CLAUDE.md generated files carry the `DO NOT MODIFY THIS FILE MANUALLY` banner, and every other generated artefact in the repo has one. `.gitattributes` only collapses the diff on github.com — in an editor, in `grep`, or in a local `git diff` there's nothing saying the file is disposable, and `rmSync` eats hand-edits without warning. A `// @generated by scripts/generate-asset-templates.mjs` line in the emitted template would cover it everywhere.

williankl (dismissed):
Approving. The two-way manifest/enum check is the right instinct — I especially like that it fails rather than emitting a broken mapping, and most of what I found below is that check working as designed.

Two things for the merge rather than the code:

- This one can't land on its own — `brand-teya-pos` is missing from the manifest, so the generator exits 1 on this branch. #383 adds it, so the stack is fine in order.
- The branch is ~5 commits behind main, which has since picked up seven icons — `Bus`, `LayDown`, `PaintRoller`, `Suitcase`, `TShirt`, `Trees`, `Wrench` — that aren't in the manifest either. The generator will fail on rebase for the same reason. Cheap to fix, just needs doing before merge.

figma/icons.manifest.json
felipeemarcon: Fixed properly in f9c7e95 rather than left to the stack: `brand-teya-pos` is now mapped here and its two templates are generated on this branch, so #382 reproduces from its own generator.

figma/icons.manifest.json
felipeemarcon: You are right, nothing read it — dropped from all three manifests in f9c7e95, so `documentUrlSubstitutions` is the only place file keys live and the README sentence is true again.

figma/scripts/generate-asset-templates.mjs
felipeemarcon: Good catch, and yes. f9c7e95 splits the run in two: every asset and platform validates first, and the script exits before the first `rmSync` if anything failed. Checked both paths — a mismatched enum and a filename collision each leave the tree untouched.

figma/scripts/generate-asset-templates.mjs
felipeemarcon: Correct on both counts. Dropped the concept in f9c7e95 — code, the empty manifest field, the README paragraph and the line in the publish skill. `knownUnmapped` covers the real case.

figma/scripts/generate-asset-templates.mjs
felipeemarcon: Both done in f9c7e95. Two names that pascal to the same file are now an error naming both entries, and every generated template carries the `DO NOT MODIFY THIS FILE MANUALLY` banner, which the CLI parses past fine — validate passes with it in place.

williankl (commented):
Re-reviewed after `f9c7e95`, and I checked each one rather than taking the reply on trust:

- `brand-teya-pos` is mapped and both its templates are generated on this branch ✅
- `fileKey` is gone from all three manifests ✅
- the generator is genuinely two-phase now — every asset and platform validates, `exit(1)`, and only then does the first `rmSync` run. That's exactly the fix ✅
- `aliases` removed from the code, the manifest, the README and the publish skill ✅
- the collision guard names both entries and the manifest to fix ✅
- the `DO NOT MODIFY THIS FILE MANUALLY` banner is on the generated templates ✅

Nice work, that was a thorough pass.

Not approving yet for one reason, and it's the drift I flagged last round rather than anything in your changes.

`icons.manifest.json` lives on this branch, and it's still missing the seven icons main picked up in #409 — `Bus`, `LayDown`, `PaintRoller`, `Suitcase`, `TShirt`, `Trees`, `Wrench`. The new workflow catches it over on #383, but it can't catch it here, because the workflow file itself only arrives in #383.

So the green tick on this PR is a gap in coverage rather than a clean bill of health — `refs/pull/382/merge` has the seven in the enum and nothing in the manifest. Worth being deliberate about it, because once #381 lands and this retargets to main, this is the one PR in the stack that could put a failing generator on main with no CI there to stop it.

Add the seven entries here and I'll approve straight away.

felipeemarcon:
Done in `5052fb1`: the seven icons from #409 are mapped here — `bus`, `lay-down`, `paint-roller`, `suitcase`, `t-shirt`, `trees` and `wrench`, node ids read from the icons library, with the fourteen generated templates alongside them.

Agreed on the coverage gap being the real point rather than the tick: the whole stack is now rebased onto current `main` (`f627186`), so this branch has exactly what `refs/pull/382/merge` has, and `node scripts/generate-asset-templates.mjs` exits 0 on it. I checked the new ids resolve by validating against Figma, and `LemonadeIcons.Bus` renders in Dev Mode on the `bus` node.

The workflow still only arrives in #383, so this PR keeps the gap you described until that lands. Worth merging the stack in order rather than retargeting this one at main on its own.

williankl (approved):
All seven icons from #409 are mapped now, with their templates generated on both platforms, so the drift that was failing the stack is gone.

Re-verified the rest of the round-one findings while I was in here, all still good: `brand-teya-pos` mapped, `fileKey` gone from all three manifests, `aliases` gone entirely, the `DO NOT MODIFY THIS FILE MANUALLY` banner on the generated templates, the two-phase generator (`if (failed) process.exit(1)` before the first `rmSync`), and the collision guard naming both entries.

Approving. Nice turnaround on all of that.

PR #383, opened 2026-09-08
feat(figma): connect components

felipeemarcon:
# Summary

Connects 38 Lemonade components to Figma Code Connect, on Compose and SwiftUI at parity. Designers' components in Dev Mode, and agents reading Figma through the MCP, now get real `LemonadeUi.*` code instead of raw layer output. The same Figma component yields Kotlin under the `Compose` label and Swift under `SwiftUI`.

ActionListItem · BlockSkeleton · BoxSelection · BrandLogo · Button · Card · Checkbox · Chip · CircleSkeleton · ContentListItem · CountryFlag · DatePicker · Divider · HistoryTimeline · Icon · IconButton · LineSkeleton · Link · ListItem · Notice · PinCode · RadioButton · ResourceListItem · SearchField · SegmentedControl · SelectField · SelectListItem · Spinner · SwipeActionRow · Switch · SymbolContainer · Tabs · Tag · TextField · TextFieldWithSelector · Tile · Toast · Tooltip

Nine internal Figma parts get their own templates so their parent can read them: Card Heading, Card Footer Action, the SegmentedControl and Tabs tab items, the History Timeline row and indicator, and the Date Picker week rows.

**Stacked on #382**, which must land first: components published without their assets render the wrong language's snippet. The Compose-only components are stacked on top of this PR, in their own PR.

## Decisions worth a look

- **Slots** render the Lemonade components they hold, read through `connectedInstances`, indented into place with their imports. A slot holding none keeps a `/* … */` placeholder. Tile's required `icon` and Chip's `leadingIcon` come from the Icon in their slots.
- **Card** takes its header and footer from the nested Card Heading and Card Footer Action, as `CardHeaderConfig` and `CardFooterActionConfig`. They only appear when the instance has them switched on.
- **TextFieldWithSelector**'s selector becomes its `leadingContent`, built from the designer's flag, icon or brand logo and the selector text.
- **DatePicker** picks `DatePicker` or `DateRangePicker` from whether the design shows a range. The months view has no code equivalent and stays unconnected.
- **PinCode**'s disabled state and **BlockSkeleton**'s Figma sizes emit a NOTE, since the code doesn't support them.
- **Skeleton** line heights are named one step apart in Figma and code (Figma's `Small` is the code's `XSmall`), so the template matches on the height token.
- **SymbolContainer**'s amber voice is still `Caution` in Figma and `Warning` in code. The template maps across it, but it's worth renaming in Figma.
- **SwiftUI** rejects trailing commas, so Swift templates put the comma before each optional argument. `TextField.input` is a `Binding`, emitted as `.constant("…")`.

## Deliberately unmapped

- `Interaction State` and `Device` everywhere: runtime state, or no code equivalent.
- `Link`'s `Show Indicator`, `Notice`'s icon swap and `SearchField`'s trailing slot: the code has nothing to map them onto.

The full list of mapping decisions lives in `figma/README.md`.

## How to test

```bash
cd figma && npm ci
FIGMA_ACCESS_TOKEN=figd_... npm run validate
```

In Dev Mode, select a Button, Tag or Card and switch the Code Connect panel between `Compose` and `SwiftUI`. The same instance should give Kotlin and Swift.

## API Dump

No public API changes.

https://claude.ai/code/session_015jzGgFffeAzM6LMp3dMVkk


figma/connect/Notice.figma.ts
```diff
@@ -0,0 +1,46 @@
...
+const title = withTitle ? instance.getString('↪ ✍️ Title') : undefined
+
+const showIcon = instance.getBoolean('◉ Show Icon')
+const actionLabel = instance.getBoolean('◉ Show Action')
+  ? instance.getString('↪ ✍️ Action Label')
+  : undefined
+
+export default {
+  example: figma.kotlin`LemonadeUi.Notice(
+    content = "${content}",
```
williankl: Nothing escapes the interpolated strings anywhere in the templates, and I think this is the one that bites on ordinary content rather than on a rename:

- `$` — "Total $5 due" emits `content = "Total $5 due"`, which Kotlin reads as a string template expression and won't compile
- `"` — `Press "Continue"` ends the literal on both platforms
- a multi-line text layer puts a raw newline inside the quotes

Notice content, Tooltip content, Toast label and the ListItem descriptions all look prone to it. A `quote()` helper in `shared/render.ts` used everywhere would cover all three, and it'd sit naturally next to `slotIcon`.

Worth saying `check-templates.mjs` structurally can't catch this one, since it never runs a template. This is the finding I'd want fixed before the first publish.

figma/connect/ActionListItem.figma.ts
```diff
@@ -0,0 +1,47 @@
...
+const showDivider = instance.getEnum('◉ Show Divider', { True: true, False: false })
+
+const leading = instance.getBoolean('◉ Show Leading')
+const trailing = instance.getBoolean('◉ Show Trailing')
+export default {
+  example: figma.kotlin`LemonadeUi.ActionListItem(
+    label = "${label}",${topLabel ? `
+    topLabel = "${topLabel}",` : ''}${supportText ? `
+    supportText = "${supportText}",` : ''}
+    onItemClicked = { },${voice !== 'Neutral' ? `
```
williankl: `undefined !== 'Neutral'` is true, so if the lookup ever fails this emits `LemonadeListItemVoice.undefined`. `HistoryItem.figma.ts:28` already guards it the right way:

```suggestion
    onItemClicked = { },${voice && voice !== 'Neutral' ? `
```

Same shape in `CardHeading.figma.ts:24` (and `:33`, where the import rides the same condition), `SelectListItem.figma.ts:38`, and the three SwiftUI twins.

figma/connect/Tile.figma.ts
```diff
@@ -0,0 +1,51 @@
...
+const disabled = instance.getEnum('◉ Is Disabled', { True: true, False: false })
+
+const topAccessory = instance.getBoolean('◉ Show Top Accessory')
+const icon = slotIcon('🧩 Leading Slot')
+
+export default {
+  example: figma.kotlin`LemonadeUi.Tile(
+    label = "${label}",${icon ? figma.kotlin`
+    icon = ${icon},` : `
+    // TODO: icon — set the LemonadeIcons entry the design uses`}
```
williankl: When `slotIcon` comes back empty this drops the argument and leaves a bare comment, but `Tile.kt:67` has `icon: LemonadeIcons` with no default and the other overload requires `leadingSlot` — so the snippet doesn't compile on that path.

`check-templates.mjs` won't catch it either: rules 3 and 4 are subset and ordering checks, not required-parameter checks.

A placeholder that compiles (`icon = LemonadeIcons.Heart, // TODO: …`) is uglier but more useful than one that doesn't, and it matches the "TODO for the developer to fill" pattern the README already describes. Same shape in `Icon`, `IconButton` and `SymbolContainer` on both platforms.

figma/connect/Switch.figma.ts
```diff
@@ -0,0 +1,37 @@
+// url=<LEMONADE_COMPONENTS>?node-id=5144-128
+// source=kmp/ui/src/commonMain/kotlin/com/teya/lemonade/Switch.kt
+// component=Switch
+import figma from 'figma'
+
+const instance = figma.selectedInstance
+
+const checked = instance.getEnum('◉ Checked', { True: true, False: false })
+const disabled = instance.getEnum('◇ Is Disabled', { True: true, False: false })
```
williankl, line 11: Three keys in this file look like outliers and I'd rather ask than assume:

- `◇ Is Disabled` here — the other 13 templates all use `◉ Is Disabled`
- `☰ Variant` on line 10 — the only `☰` in the PR, and `Checkbox.figma.ts:16` uses `◇ Layout` for the same Standalone/Content axis
- `↪ ◉ Show Description` on line 13 vs `Checkbox.figma.ts:19`'s `↪ Show Description`

If any of them is wrong it degrades to `undefined` silently, which is the failure mode the README calls out. `ContentListItem.figma.ts:23-27` has the same smell (`◉ Show divider`, lowercase d, and two unprefixed `Show Leading`/`Show Trailing`).

Worth a Dev Mode spot-check — or am I misreading the Figma side?

figma/scripts/check-templates.mjs
```diff
@@ -0,0 +1,174 @@
...
+// 3. Compose named arguments exist on some overload of the function called.
+const kotlinDecls = declarations(
+  kotlinSources,
+  /^(?:public\s+)?(?:@\w+\s+)*(?:fun\s+(?:<[^>]*>\s*)?(?:LemonadeUi\.)?|(?:data\s+)?class\s+)(?<name>\w+)(?:<[^>]*>)?\s*\(/gm,
+  (p) => p.match(/^(?:va[lr]\s+)?(\w+)\s*:/)?.[1],
+)
+for (const file of [...compose, ...shared]) {
+  for (const call of calls(readFileSync(file, 'utf8'), '=')) {
+    const overloads = kotlinDecls.get(call.name)
+    if (!overloads) continue
```
williankl: `if (!overloads) continue` means a snippet calling something that doesn't exist — a typo, or a component renamed on the Kotlin side — gets skipped rather than reported, which is probably the most likely way this check quietly stops working. Same at line 163, and line 96's `if (source && …)` lets a template with no `// source=` header through entirely.

Nice script otherwise — automating the Swift declaration-order rule is a good call, that's not a fun one to catch by eye.

figma/package.json
```diff
@@ -4,6 +4,7 @@
   "version": "0.0.0",
   "description": "Figma Code Connect mappings for the Lemonade design system.",
   "scripts": {
+    "check": "node scripts/check-templates.mjs",
```
williankl, line 7: `check` needs no token and runs offline in seconds, but nothing runs it — there's no Figma job in `.github/workflows/`. `token_drift.yml` is almost exactly the shape this wants: `pull_request` with a `paths:` filter, no secrets, regenerate-and-fail-on-drift. It also has no `branches:` filter, so unlike `kmp_ci.yml` it would actually run on the stacked PRs.

One subtlety if you do wire it: `git diff --exit-code` alone isn't enough. When the generator fails it `continue`s without writing, so the tree stays clean while the script exits 1 — the job wants to check the exit code too.

And I'd think twice about putting `FIGMA_ACCESS_TOKEN` in Actions secrets on a public repo. `validate` needs it, `check` doesn't, and the tokenless one is the one actually worth having in CI.

williankl (dismissed):
Approving. This is a lot of careful work and the mapping decisions in the README are genuinely useful — the Caution/Warning and skeleton-height notes are exactly the things that would otherwise get "fixed" by the next person through.

Nothing here blocks the merge: none of it ships to consumers, and the templates do nothing until someone runs `publish`. The escaping one is the only finding I'd want in before a publish rather than before a merge, because it fires on ordinary designer copy rather than on a rename.

One for the design team rather than this PR — given how `getEnum` degrades, the Caution → Warning rename in Figma is worth a heads-up before anyone does it.

figma/connect/Notice.figma.ts
felipeemarcon: Agreed, this was the one worth blocking on. 43d3832 adds `quote()` to `shared/render.ts` and routes every string read from Figma through it — quotes, backslashes and newlines on both platforms, plus `$` on Kotlin only, since Swift has no `\$` escape. The helper picks the language from the tag. You are right that `check-templates` cannot see this; it is noted as a limit in the README.

figma/connect/ActionListItem.figma.ts
felipeemarcon: Fixed in 43d3832: ActionListItem, CardHeading and SelectListItem now check the value first, on both platforms, matching HistoryItem. CardHeading's conditional import rides the same guard.

figma/connect/Tile.figma.ts
felipeemarcon: Right, and it applied to Icon, IconButton and SymbolContainer too. 43d3832 emits the argument with a placeholder plus a `// NOTE:` saying the icon did not resolve, so the snippet compiles and still says something is wrong.

figma/connect/Switch.figma.ts
felipeemarcon: Checked all of them in Dev Mode rather than assuming — the templates match Figma exactly. Switch really does use `◇ Is Disabled`, `☰ Variant` and `↪ ◉ Show Description`, and Content List Item really does use unprefixed `Show Leading`/`Show Trailing` and lowercase `◉ Show divider`. The library is inconsistent; the templates follow it. Worth raising with design, but not a template fix.

figma/scripts/check-templates.mjs
felipeemarcon: Fixed in 43d3832. A `LemonadeUi.X` call naming nothing in the sources is now reported on both platforms, and a template with no `// source=` header fails instead of being skipped. Bare calls stay skipped on purpose, so `Row` and `HStack` do not become false positives.

figma/package.json
felipeemarcon: Wired in ea62664: `.github/workflows/figma_code_connect.yml`, modelled on `token_drift.yml`, `pull_request` with a `paths:` filter and no secrets. It runs `check-templates.mjs` and then regenerates the assets, and as you said it checks the exit code, not just a clean tree. Agreed on the token: `validate` stays local.

williankl (commented):
Re-reviewed both new commits.

The escaping work is right, and I checked the order rather than eyeballing it: backslash first, then quotes, then newlines, with `$` last and Kotlin-only. That ordering matters and it's correct — the backslashes the later passes introduce don't get double-escaped. `\(` in Swift is neutralised by the backslash pass too, so interpolation can't sneak in on that side either. Routing it through `renderer` so it picks the language off the tag is neat.

Also confirmed: the seven `!==` guards all check the value first now, on both platforms; `Tile`, `Icon`, `IconButton` and `SymbolContainer` always emit the required argument; and `check-templates.mjs` reports unknown names and fails a template with no `// source=`.

The workflow is a good build — no secrets, a `paths:` filter, no `branches:` filter so it actually runs on the stacked PRs, and it checks the exit code rather than just a clean tree. The comment explaining why `validate` stays local is the right call for a public repo.

Holding the approval only because it's red, and it's red for a real reason: the seven icons from #409 still aren't in the manifest. Your new check caught it on its first run, which is a decent endorsement of the check. The fix belongs on #382 since that's where `icons.manifest.json` lives — I've commented over there.

One small thing while you're in here. In `shared/render.ts`:

```js
const kotlin = tag``.language === 'kotlin'
```

If `.language` ever comes back undefined, `$` escaping silently switches off for Kotlin and we're back to the bug you just fixed, with nothing to notice. Is that a documented property on the tag result, or worth asserting on?

felipeemarcon:
Both points from the re-review are done, and CI is green on `69239d7`.

**The seven icons.** Fixed on #382 as you said, in `5052fb1` — `Bus`, `LayDown`, `PaintRoller`, `Suitcase`, `TShirt`, `Trees` and `Wrench` are mapped from the icons library, and the whole stack is rebased onto current main so the branches and `refs/pull/*/merge` agree. That also means #382 now regenerates cleanly on its own.

**`tag``.language`.** Fair challenge, and it was a real silent-failure path: anything that was not `kotlin` fell through to the Swift branch. `69239d7` reads the language once and throws when it is neither `kotlin` nor `swift`, so a change in the CLI fails the publish instead of quietly dropping the `$` escape. I confirmed both values in Dev Mode rather than trusting the types — `figma.kotlin` reports `"kotlin"`, `figma.swift` reports `"swift"` — and the skill's `references/template-api.md` now records that, with the reason.

One thing to flag: #407 is closed as merged, wrongly. While rebasing this stack I pushed that branch pointing at its own base for a moment, and GitHub read that as merged. Nothing actually landed — `feat/figma-connect-components` and `main` are both untouched. It continues as #415 with the same branch and content.

williankl (commented):
CI is green on `fe8a17c8` and the `.language` guard is exactly right — throwing on an unknown language rather than silently falling back to no `$` escaping is the safe direction, and documenting it in `template-api.md` is a nice touch.

One thing to flag rather than a new finding: merging #407 into this branch carried its open comment across with it. `figma/connect/DropdownItem.figma.ts:15` is still interpolating raw, and it's now this PR's to fix:

```js
text = "${instance.getString('✍️ Label')}"
```

I re-swept every template on `fe8a17c8` and it's the only one left — everything else in `connect/`, `connect-swiftui/` and `shared/` routes through `quote()`. It propagates too, since `Dropdown` inlines the child's snippet through `executeTemplate()`, so a `$` or a `"` in an item label takes the parent down with it.

Worth being aware that neither gate catches this one: `check-templates.mjs` never runs a template, and the new workflow only checks that snippets compile against the sources and that the assets match. So it'll stay invisible until someone publishes and a designer writes a price into a dropdown.

Destructure `quote` off `renderer(instance, figma.kotlin)` and wrap the `getString`, same as everywhere else, and I'll approve. That's genuinely the last thing on my list for this stack.

figma/connect/DropdownItem.figma.ts
```diff
@@ -0,0 +1,25 @@
...
+const { quote } = renderer(instance, figma.kotlin)
+
+const holder = instance.getBoolean('◉ Show Leading') ? instance.findInstance('Icon') : undefined
+const glyph = holder && holder.type === 'INSTANCE' ? holder.getInstanceSwap('🧩 Icon') : undefined
+const icon = glyph && glyph.type === 'INSTANCE' ? glyph.executeTemplate().example : undefined
+const trailing = instance.getBoolean('◉ Show Trailing')
+const supportText = instance.getBoolean('◉ Show Support Text')
+
+export default {
+  example: figma.kotlin`LemonadeUi.DropdownItem(text = "${quote(instance.getString('✍️ Label'))}", onClick = { }${icon ? figma.kotlin`, leadingIcon = ${icon}` : ''}${trailing ? ', trailingSlot = { /* trailing content */ }' : ''})`,
```
felipeemarcon, line 17: Fixed in `941cd09` — `quote` destructured off `renderer(instance, figma.kotlin)` and wrapped around the `getString`, same as everywhere else.

You were right about how it survived: the escaping pass ran on this branch while `DropdownItem` was still over on the Compose-only branch, so the sweep never saw it. I re-swept `fe8a17c8` the same way after fixing it and it is now zero — every interpolation inside a string literal across `connect/`, `connect-swiftui/` and `shared/` goes through `quote()`.

Published and checked the parent too, since that was the propagation path: the Dropdown snippet still renders its four items with their icons.

And agreed on the gate gap. Neither `check-templates.mjs` nor the workflow can see this class, because catching it means executing a template against a design rather than reading source. The README already records that limit for `check`; if this happens again the honest fix is a runtime check against a fixture rather than more static analysis.

williankl (dismissed):
That's the last one — approving.

Re-swept every template on `941cd096` and there are now zero unescaped reads: everything in `connect/`, `connect-swiftui/` and `shared/` routes Figma text through `quote()`. Both checks are green.

Also re-verified the earlier findings survived the #407 merge and the rebases: all the enum guards check the value before comparing, `Tile` always emits its required `icon` with a NOTE when the design's doesn't resolve, `nestable: false` is on the stubs and containers, and the `.language` guard is in place.

Good stack. The two-phase generator and the `paths:`-filtered tokenless workflow are the parts I'd point someone at — between them the drift that bit this three times can't reach main unnoticed again.

Two things worth carrying forward rather than reopening here:

- `check-templates.mjs` reads template source rather than running templates, so the escaping class of bug is invisible to it. Worth remembering when the next template lands.
- Give design a heads-up before the SymbolContainer `Caution` → `Warning` rename in Figma, since that's exactly the change `getEnum` degrades to `undefined` on.

One process note, not blocking: #407 was merged into this branch while it still had an open comment and `REVIEW_REQUIRED`. It worked out because I re-checked and the finding moved here with it, but branch protection only guards `main`, so the stacked branches will let that through. Requiring review on them too would close it off if you think it's worth the friction.

williankl (approved):

PR #401, opened 2026-09-11
feat(symbol-container): add a theme overload to the KMP Symbol Container

felipeemarcon:
# Summary

Symbol Container can now be styled by a themed palette on KMP, solid or subtle. Stacked on #400.

Why: to match the new `Symbol Container (Themed)` component in Figma, for colour that carries meaning the voices don't cover, such as categories or per-role accents.

- New icon and text overloads that take a required `theme: ThemedStyle` and no `voice`, so the two can't be combined. Existing `voice` calls are unaffected.
- `ThemedHue.Violet` uses the hue's solid palette and `ThemedHue.Violet.subtle` its subtle one: `background` for the fill, `border` for the stroke, `onBackground` for the icon or text.
- Marked `@ExperimentalLemonadeApi`, since they build on the themed layer.
- Internally, the shared surface now takes resolved colours instead of a voice, so the voice and theme overloads share one icon helper and one text helper.
- detekt's `TooManyFunctions` no longer counts `@Composable` functions, set in `kmp/config/detekt/detekt.yml` the same way `LongParameterList` already ignores them. That makes the file-level suppressions in `Text.kt`, `IconButton.kt`, `ListItem.kt` and `SwipeActionRow.kt` unnecessary, so they are removed. `InlineCalendar.kt` keeps its own, because its 15 functions are plain date helpers.
- A themed preview, and Themes (Solid) and Themes (Subtle) sections in the sample's Symbol Container page.

```kotlin
LemonadeUi.SymbolContainer(icon = LemonadeIcons.Heart, contentDescription = "Favourite", theme = ThemedHue.Violet)
LemonadeUi.SymbolContainer(icon = LemonadeIcons.Heart, contentDescription = "Favourite", theme = ThemedHue.Violet.subtle)
```

### How to test

1. `cd kmp && ./gradlew :composeApp:run`, open **SymbolContainer**, scroll to **Themes (Solid)** and **Themes (Subtle)**. Toggle light and dark.
2. `./gradlew :ui:desktopTest --tests 'com.teya.lemonade.SymbolContainerTest'`.

## Figma component

https://www.figma.com/design/91S16rhVrl5wivqV66fNjm/?node-id=21779-772 (`<Hue>` solid, `<Hue> Subtle` subtle)

## Screenshot

See the Themes sections of the sample app.

## API Dump

**Verdict:** ADDITIONS_ONLY

- Two new `SymbolContainer` overloads (icon and text, each taking `ThemedStyle`) in `ui`. They are new functions; the existing overloads' signatures are unchanged.

**Genuine breaking changes (if any):** none.


kmp/ui/src/commonMain/kotlin/com/teya/lemonade/SymbolContainer.kt
```diff
@@ -1,10 +1,14 @@
+@file:Suppress("TooManyFunctions")
```
williankl: can we move this suppression to the build.gradle file instead? doesn't make sense to have this limitation in a Jetpack Compose first repo

williankl (changes requested):
one suggestion

kmp/ui/src/commonMain/kotlin/com/teya/lemonade/SymbolContainer.kt
felipeemarcon: Moved it to `kmp/config/detekt/detekt.yml` (the config the build points detekt at): `TooManyFunctions` now ignores `@Composable` functions, the same way `LongParameterList` already does. That also let me drop the suppression from `Text.kt`, `IconButton.kt`, `ListItem.kt` and `SwipeActionRow.kt`. `InlineCalendar.kt` keeps its suppression, because it goes over the limit on plain date helpers, not composables.

williankl (approved):

PR #406, opened 2026-09-13
fix(topbar): keep derived-state reads out of the collapsing title's layer block

RomanTruba-Teya:
## Problem
`LemonadeUi.TopBar`'s large title can render clipped to another element's shape. In the Teya app's settlements Review sheet it takes the clip of a text link below it (a different link on each render), and a `startCollapsed` bar also loses its collapsed offset.

The collapsing title's `graphicsLayer` block reads `collapseProgress`, a `derivedStateOf`. When it recalculates inside the block, Compose delivers pending snapshot notifications mid-block, another node's layer update overwrites the layer scope all nodes share, and the title applies that node's clip. Reproduces on Compose UI 1.11.2 and 1.12.1.

## Solution
Read plain state in the layout and layer phases:
- the scroll `Animatable` is held in a `mutableStateOf` instead of a `derivedStateOf`
- `heightOffset` is a plain getter, and the layer block reads a new internal `currentCollapseProgress`
- the public `collapseProgress` and `isCollapsed` are unchanged; no public API changes (`NO_CHANGES`)

## Test plan
- Robolectric repro in the Teya app (TopBar next to a sibling that clips): the title layer applies `clip=false` and the correct collapsed offset; fails on 0.40.0
- Settlements Review screenshots against a local build: 4/50 flaky before, 0/50 over 5 runs after; other goldens unchanged
- `ktlintCheck`, `apiCheck`, `detektMetadataCommonMain` pass
- Manually tested various screens and scenarios, no visual regression

## Screenshots
SettlementsReviewFrequencyChainsawShowcasePreviews_BaActivationEmpty_Dark before-after
SettlementsReviewFrequencyChainsawShowcasePreviews_BaActivationEmpty_Light before-after


felipeemarcon (approved):
LGTM

williankl (approved):

PR #407, opened 2026-09-14
feat(figma): connect the Compose-only components

felipeemarcon:
# Summary

Connects Lemonade's seven Compose-only components to Figma Code Connect: Dropdown, DropdownItem, BottomTabBar, BottomSheet, TimePicker, TimeInput and TimePickerDialog. Designers' Android designs now give real `LemonadeUi.*` Kotlin in Dev Mode and to MCP-driven agents, instead of raw layer output.

SwiftUI has no equivalent for any of them, so under the `SwiftUI` label each shows a one-line NOTE saying so. Without it, Figma falls back to another label's snippet and shows iOS developers Kotlin.

**Stacked on #383**, which must land first.

## Decisions worth a look

- **Dropdown** flattens Figma's sections into one list of `DropdownItem`s, since the code has no sections. A design with section headings, dividers or item support text gets a NOTE.
- **BottomTabBar** reads the tabs and the selected one from the bar's slot. The tab badge has no code equivalent, and a solid icon on the selected tab isn't split out into `selectedIcon`.
- **Time pickers** open with the `rememberLemonadeTimePickerState` call the component needs, seeded with the designed time. A 12-hour design is converted to the code's 0–23 hour through its AM/PM toggle. The dialog's two accessibility labels have nothing in the design to read, so they're emitted as English text.
- **BottomSheet** maps both Figma sets: Full Screen adds `skipPartiallyExpanded = true`. The code sheet has no header, so the title goes into the content as a `Text`, as in the component's KDoc. The actions slot renders its real buttons.
- `findLayers` with `traverseInstances: true` returns layers last to first, so templates that list children in order search one level at a time.

## Figma component

Pages in the components library: Dropdown, Bottom Tab Bar, Sheet, and the Android sets on the Date Picker page.

## How to test

In Dev Mode, select a Dropdown or a Bottom Tab Bar and check that the `Compose` panel lists the items in the order the design shows them, and that the `SwiftUI` panel shows the NOTE.

## API Dump

No public API changes.

https://claude.ai/code/session_015jzGgFffeAzM6LMp3dMVkk


figma/connect/BottomTabBar.figma.ts
```diff
@@ -0,0 +1,38 @@
...
+)
+
+let items = figma.kotlin``
+let selectedIndex = 0
+tabs.forEach((tab, index) => {
+  if (tab.getEnum('◉ Is Selected', { True: true, False: false })) selectedIndex = index
+  const glyph = tab.getInstanceSwap('🧩 Icon')
+  const icon = glyph && glyph.type === 'INSTANCE' ? glyph.executeTemplate().example : undefined
+  items = figma.kotlin`${items}
+        BottomTabBarItem(label = "${tab.getString('✍️ Label')}"${icon ? figma.kotlin`, icon = ${icon}` : ''}),`
```
williankl: Two things on this line.

If the instance swap doesn't resolve, this emits `BottomTabBarItem(label = "Home")` — and `icon` has no default in `BottomTabBar.kt`, so it won't compile. A fallback or a NOTE would be safer than dropping the argument.

And since each tab's own `🧩 Icon` is read, the designer's selected tab comes out as `icon = LemonadeIcons.WalletSolid` with no `selectedIcon` — the KDoc's intent is `icon = Wallet, selectedIcon = WalletSolid`. It looks right in the selected state and goes wrong the moment you tap another tab, which is the kind of snippet that teaches the wrong API.

figma/connect-swiftui/BottomTabBar.figma.ts
```diff
@@ -0,0 +1,10 @@
+// url=<LEMONADE_COMPONENTS>?node-id=17947-25065
+// source=kmp/expressive/src/commonMain/kotlin/com/teya/lemonade/BottomTabBar.kt
+// component=BottomTabBar
+import figma from 'figma'
+
+export default {
+  example: figma.swift`// NOTE: BottomTabBar is Compose-only; Lemonade has no SwiftUI BottomTabBar`,
+  id: 'bottom-tab-bar',
+  metadata: { nestable: true },
```
williankl: The NOTE stub itself is the right call — without it the SwiftUI panel falls back to the Kotlin snippet, which is worse.

But `nestable: true` means a parent Swift snippet would inline `// NOTE: BottomTabBar is Compose-only…` into the middle of a call and the component just disappears, rather than leaving the parent's `/* … */` placeholder. `nestable: false` instead?

Same question for the other NOTE stubs and for the new containers (`Dropdown`, `BottomSheetDynamic/FullScreen`, `TimePickerDialogDial/Input`) — `Card`, `DatePicker` and `SwipeActionRow` all declare `false`.

figma/shared/time.ts
```diff
@@ -0,0 +1,65 @@
...
+  const hour = value(hourField)
+  return {
+    hour: !is24Hour && hour <= 12 ? (pm ? (hour % 12) + 12 : hour % 12) : hour,
+    minute: value(minuteField),
+  }
+}
+
+// The state every time picker composable takes, seeded with the designed time.
+export const timeState = (instance, holder = instance) => {
+  const is24Hour = instance.getEnum('◇ Format', { '12 hour': false, '24 hour': true })
```
williankl: This one goes straight into a Kotlin literal, so a renamed `◇ Format` gives `is24Hour = undefined,` — and it also makes `!is24Hour` truthy on line 19, so the hour silently takes the 12-hour conversion branch as well. Two failures out of one rename.

```suggestion
  const is24Hour = instance.getEnum('◇ Format', { '12 hour': false, '24 hour': true }) ?? false
```

figma/shared/time.ts
```diff
@@ -0,0 +1,65 @@
...
+
+// The code takes a 0-23 hour; Figma shows a 12-hour clock with an AM/PM toggle.
+const readTime = (holder, is24Hour) => {
+  const layers = holder.findLayers(
+    (node) => node.type === 'INSTANCE' && (FIELDS.includes(node.name) || node.name.startsWith(PERIOD)),
+  )
+  const [hourField, minuteField] = layers.filter((node) => FIELDS.includes(node.name))
+  const period = layers.find((node) => node.name.startsWith(PERIOD))
+  const value = (field) => (field ? parseInt(field.getString('Time selector label text'), 10) || 0 : 0)
+  const pm = period ? period.getEnum('AM/PM', { True: true, False: false }) : false
```
williankl: Mapping an AM/PM property onto `True`/`False` variant names is unusual enough that I'd like to check it rather than assume — if the variants are actually named `AM`/`PM`, this returns `undefined`, `pm` is falsy, and every PM design quietly loses 12 hours.

Does Figma really spell them True/False here?

figma/shared/time.ts
```diff
@@ -0,0 +1,65 @@
...
+export const timePickerDialog = (instance, inputMode) => {
+  const buttons = instance.findLayers((node) => node.type === 'INSTANCE' && node.name === 'Primary button')
+  const label = (variant, fallback) => {
+    const button = buttons.find(
+      (b) => b.getEnum('◇ Variant', { Primary: 'Primary', Neutral: 'Neutral' }) === variant,
+    )
+    return button ? button.getString('✍️ Label') : fallback
+  }
+  return {
+    example: figma.kotlin`${timeState(instance)}
```
williankl: `TimePicker.figma.ts:8-9` deliberately digs out the nested `Android / Time Input` before reading the fields, but this passes only `instance`, so `holder` defaults to the dialog itself. `readTime` then calls `findLayers` without `traverseInstances`, which doesn't cross the instance boundary — the same reason `Dropdown` walks sections and items in two passes.

If I've got that right, every dialog snippet comes out `initialHour = 0, initialMinute = 0`, while the README says it's seeded with the designed time. Worth a look in Dev Mode — I could easily be wrong about the traversal.

williankl (dismissed):
Approving. The one-line NOTE stubs for the Compose-only components are a neat answer to the fallback trap — I checked `swiftui/Sources/` and all seven really are missing on that side, so the reasoning holds.

Comments are a couple of real bugs in the BottomTabBar mapping plus two Figma-side questions I couldn't answer from the repo. None of it blocks, and nothing here is visible to anyone until a publish.

figma/connect/BottomTabBar.figma.ts
felipeemarcon: Both fixed in 4a9c64a. An unresolved icon now emits a placeholder plus a NOTE instead of dropping a required argument. On `selectedIcon`, the design carries one icon per tab and the outline counterpart is not in it, so the snippet now carries a NOTE saying to pair the outline icon with its solid variant — better than inventing the pair or silently presenting the solid one as the default.

figma/connect-swiftui/BottomTabBar.figma.ts
felipeemarcon: Agreed. 4a9c64a sets `nestable: false` on all nine NOTE stubs, and on the containers you listed — Dropdown, both bottom sheets, both dialogs, TimePicker, TimeInput and BottomTabBar. `DropdownItem` stays nestable, since Dropdown renders it.

figma/shared/time.ts
felipeemarcon: Fixed as suggested in 4a9c64a — `?? false`. You are right that it was two failures from one rename.

figma/shared/time.ts
felipeemarcon: Checked in Figma: the Period Selector really is a boolean property named `AM/PM` with `True`/`False` variants, where True is PM. Odd naming, but the mapping is correct as written.

figma/shared/time.ts
felipeemarcon: Checked this one in Dev Mode — the dialog snippets do come out seeded: `initialHour = 20` on the 24-hour variants and `= 7` on the 12-hour ones. The dial and input fields are direct children of the dialog variant rather than inside a nested instance, so `findLayers` reaches them without traversal. `TimePicker` is the one that needs the nested `Android / Time Input`, which is why it passes a holder.

figma/connect/DropdownItem.figma.ts
```diff
@@ -0,0 +1,23 @@
...
+const instance = figma.selectedInstance
+
+const holder = instance.getBoolean('◉ Show Leading') ? instance.findInstance('Icon') : undefined
+const glyph = holder && holder.type === 'INSTANCE' ? holder.getInstanceSwap('🧩 Icon') : undefined
+const icon = glyph && glyph.type === 'INSTANCE' ? glyph.executeTemplate().example : undefined
+const trailing = instance.getBoolean('◉ Show Trailing')
+const supportText = instance.getBoolean('◉ Show Support Text')
+
+export default {
+  example: figma.kotlin`LemonadeUi.DropdownItem(text = "${instance.getString('✍️ Label')}", onClick = { }${icon ? figma.kotlin`, leadingIcon = ${icon}` : ''}${trailing ? ', trailingSlot = { /* trailing content */ }' : ''})`,
```
williankl: This one slipped through the escaping sweep — `DropdownItem` landed in this PR rather than #383, so `43d3832` didn't reach it:

```
text = "${instance.getString('✍️ Label')}"
```

A dropdown item label with a `$` or a `"` in it breaks the snippet, and it propagates — `Dropdown` inlines the child's output through `executeTemplate()`, so the parent breaks along with it.

I swept every template in `connect/` and `connect-swiftui/` and this is the only one left; `shared/time.ts` and `shared/bottom-sheet.ts` both route through `quote()` correctly.

Same shape as everywhere else should do it — destructure `quote` off `renderer(instance, figma.kotlin)` and wrap the `getString`.

williankl (commented):
Re-reviewed `4a9c64a`. `nestable: false` is on all nine stubs and every container you listed, `?? false` is on `is24Hour`, and the BottomTabBar icon always emits now with a NOTE rather than dropping a required argument. The `selectedIcon` NOTE is the right answer too — better than inventing the outline/solid pair or quietly presenting the solid one as the default.

Thanks for actually checking the three Figma questions in Dev Mode instead of waving them through. The `AM/PM` True/False naming and the dialog reaching its fields without traversal both hold up with the concrete values you quoted, and the TimePicker/dialog asymmetry makes sense now.

One the escaping sweep missed — inline comment below. Holding the approval on the red CI, which is the manifest drift rather than anything in this PR (commented on #382, that's where the fix lands). Get that green and fix the one line here and I'll approve both.

felipeemarcon:
Closing note for anyone landing here from the stack: GitHub marked this PR merged by mistake. During a rebase I pushed this branch pointing at its own base for a moment, which GitHub read as "every commit is in the base". Nothing was merged — the five commits are still only on `feat/figma-connect-compose-only`, and neither `feat/figma-connect-components` nor `main` contains them.

The work continues in #415, with the same branch and the same content. The approval and the four review threads here still apply; sorry for the split history.

PR #414, opened 2026-09-17
fix(topbar): pad only the display cutout above and beside the bar

RomanTruba-Teya:
## Problem
`LemonadeUi.TopBar` grows a band below its content when a device is turned with its camera at the bottom edge, on a foldable or a tablet in landscape. In the Teya app the Money and Sales tab bars and every details screen's top bar showed it on a Pixel 9 Pro Fold rotated 180°.

`CoreTopBar` applies `displayCutoutPadding()`, which pads the cutout on every side. With the camera at the bottom the cutout inset sits on the bar's bottom edge, so the bar pads it below its content.

## Solution
`CoreTopBar` pads only the top and horizontal display cutout: `windowInsetsPadding(WindowInsets.displayCutout.only(Top + Horizontal))`. The status bar padding is unchanged.

## API Dump
No public API changes (`NO_CHANGES`).

## Test plan
- In the Teya app on the Pixel 9 Pro Fold emulator with the camera at the bottom edge: the Money, Sales and details top bars grew below their content; consuming the bottom cutout before the bar, which leaves this padding with only the top and side cutout, removed the band on every screen
- `ktlintCheck`, `apiCheck` and `detektMetadataCommonMain` pass; `:ui` compiles for Android and the iOS simulator
- `bcv-check.sh --ci`: `NO_CHANGES`


williankl (approved):
very good catch 🚀

PR #416, opened 2026-09-17
fix(topbar): give a filled toolbar menu the circle a filled button gets

Saed-Eltalwy:
## The problem

`.filled` is one prominence shared by `ToolbarIconButton` and `ToolbarIconMenu`, but the two disagree on what it looks like: the **button is a circle, the menu is an oval**.

Both go through the same line:

```swift
case let .filled(tint):
    self.buttonStyle(.borderedProminent).tint(tint)
```

## Why

SwiftUI promotes a `ToolbarItem` to a real `UIBarButtonItem` only when it recognises the button's label — the bridge `LemonadeToolbarIconButton.swift` already documents. A `Menu` is not recognised, so it crosses as a plain view, and there `.borderedProminent` sizes a capsule around the glyph instead of taking the bar item's square.

`buttonBorderShape(.circle)` does not survive the trip either; I tried it first. Pinning the frame to the bar item's own diameter (`size900`, 36) is what actually lands it on the circle.

A plain menu is untouched — it draws no background to reshape.

## Found via

Teya's Sales and Money tabs, where the `+` reads as an oval next to circular buttons elsewhere in the app. The consumer-side workaround is https://github.com/saltpay/teya-business-app/pull/4699, which I would like to delete once this ships — a consumer passes only an icon, a description and a prominence, so it has no lever for this and has to reach around the component to correct it.

## Sample app

Nothing in `SampleApp` used `filled(tint:)`, which is how the two components came to disagree unnoticed. **TopBar → Basic with Filled Actions** now puts a filled button beside a filled menu in one bar, so a difference in shape is obvious at a glance.

## Testing
|after | before |
|-|-|
| image | image |

`swift build` passes, including the macOS 12 target this file also compiles for. `LemonadeSampleApp` builds and runs on the simulator, and the new demo was checked both with the fix in and with it backed out.

## Open question

A reviewer on the consumer PR reported not seeing the oval, so it may depend on iOS version or device. There is no version branch in this file, so any difference comes from SwiftUI's own bar-item bridging, which changed in iOS 26. Worth a second pair of eyes on an older OS before merge — it decides whether the frame should be applied unconditionally or gated.


swiftui/Sources/Lemonade/Components/LemonadeToolbarIconButton.swift
```diff
@@ -131,6 +133,31 @@ private struct LemonadeToolbarIconMenuView<Content: View>: View {
...
+
+    @ViewBuilder
+    func body(content: Content) -> some View {
+        switch prominence {
+        case .plain:
+            // Nothing to reshape: a plain menu draws no background.
+            content
+        case .filled:
+            content
+                .frame(width: LemonadeSizes.size900.value, height: LemonadeSizes.size900.value)
```
felipeemarcon, line 155: Not sure if we should force the width and height light this. I'm concerned if other Apple's devices don't use different size for the topbar buttons, since we use the full-native ones for navigation (back and close). It could decompensate the topbar visually.

